Issue #5, opened 2017-04-13
Support for circular dependency checks


marcghorayeb:
Hello,
👍 for the great repo. I saw the roadmap and was wondering if you had progressed on the circular dependency check?
Thanks in advance,

sverweij:
Thanks for the kind words 😊 

I haven't started on a circular dependency check yet.

marcghorayeb:
I work on a large codebase mixing up JS/Coffee/TS and we are starting to see a lot of problems due to circular dependencies, making a tool to check these quite important to us.
I haven't any clue as to how complicated it would be to develop such a check but I'm ready to help out if you're open to PRs?

sverweij:
Yesterday I've given some thoughts on how to implement this. Today I have a version running on my machine that correctly detects circular dependencies. It needs some polishing; I will try to publish coming week. 

Are you willing to beta test it on your code base and provide me with feedback?

(Generally PR's are welcome b.t.w.)

marcghorayeb:
That was quick! Thanks a lot.
Sure, will check out the PR against our codebase, anything you want me to test out first?

sverweij:
Cool!

Nothing specific to test first - just want to know whether it works as you expect it to (accuracy, usability, speed). 

B.t.w. how do (/ will) you use dependency-cruiser in your development process?

marcghorayeb:
Did a quick check, it seems to work as expected 👍  I'll do some more testing tomorrow.

At first the accuracy seems good. Maybe some more info would be nice when outputting to the console (what path was used to see the circular?). The speed seems acceptable, we have around 2000 files and it went through it in approximately 10 seconds (no transpilers, going through the transpiled code, so only JS files in it).

At first we only want to show some warnings when the branches are being built on our CI tool (Atlassian Bamboo) and later on we might enforce no circular dependencies. At the very least, we know of a couple of files that are quite sensitive to any dependency changes and we might enforce no more than X circulars on a given file.

sverweij:
Thanks. This is useful feedback!

- I'll see if I can add the offending path to the output - doesn't seem to be very complex to do and it'd be massively useful. Thanks for that suggestion! I'll include it with the publish later this week if possible.
- It might be possible to help you enforce to not have > X circulars on a file - but I'll have to think about how to implement it.

[...]

PR #6, opened 2017-04-15
Detect circular dependencies

sverweij:
This will add a circular dependency check to the arsenal of dependency-cruiser validations.

The feature has been on the roadmap quite some time - bumped & implemented it because of real life demand (#5, @marcghorayeb)

### usage
- Add a rule to your .dependency-cruiser.json like this:
```json
{
        "name": "no-circular",
        "severity": "warn",
        "comment": "Warn in case there's circular dependencies",
        "from": {},
        "to": { "circular": true }
}
```
For each dependency this will answer the question: _When I follow this dependency, will I end up with myself?_ and hoist a warning flag when it is.

- Run dependency-cruiser 
```
depcruise -v -- src
```

### Example: the coffee-script compiler source
Running with this rule e.g. against the coffeescript transpiler code base will yield
```
  warn no-circular: coffee-script.coffee → register.coffee
  warn no-circular: register.coffee → coffee-script.coffee

✖ 2 dependency violations (0 errors, 2 warnings). 23 modules cruised.
```

or, rendered as a graph (`depcruise -v -x node_modules -T dot  src | dot -T svg > coffee-deps.svg`)
coffee-deps

### changes
- adds a `circular` attribute the json schema for .dependency-cruiser.json configuration files.
- adds a `circular` attribute to the json schema for the output format
- adds some code to check for circular dependencies
- moves the validations from _during_ the extraction to _after_ the extraction (just after when the time circular dependencies can be determined).
- adds the circular check to the `--init-rules` configuration with a _warning_ level.   
Although circular dependencies are a code-smell, software usually works correctly, and sometimes it's intended (ref CoffeeScript) - so _warn_ is a safe default. In real life projects I'd probably set it to _error_ so I'd get flags in my ci and/ or pre-push.
- updates the rules reference to include the circular rule


marcghorayeb:
👍

Issue #32, opened 2018-02-16
Use glob for CLI arguments


ajafff:
Currently the CLI doesn't support globbing. That's not a problem on linux shells as they already expend glob patterns.
If I wanted to use the same command on Windows however, I get an error like `ERROR: Can't open 'packages/*/src' for reading. Does it exist?`.
Because of this, I cannot really use this package on a windows CI to test my dependencies.

It would be great if you could consider using `glob` to find the files.

As a side effect that also allows using extglob and globstar features even if the shell doesn't support that. You only need to wrap the glob pattern in quotes and let depcruise do the globbing instead of the shell.

sverweij:
Hi @ajafff first of: thanks for reporting this and the other three issues - the level of details is excellent to work with! I'll respond to them one by one in context, but here's some stuff that goes for all 4:

### ordering
I can't pick them all up at the same time I'm going to need to order them. This is my working assumption - please correct me:
1. #32 (Use glob for CLI arguments) - because that is an issue that prevents cross platform use of arguments that in some scenario's is blocking. 
2. #33 (resolve symlinks) - because it's a bug, 
3. #34 (introduce a severity level for the 'allowed' set of rules) - new feature 
4. #35 (use closest package.json) - improve support for mono repos

(I don't guarantee I'll pick em up in that order; if there's low hanging fruit that takes only an hour to fix that might get attention first). 

That said - a helping hand is welcome.

### releases
Once an issue is fixed I'll release it. First as a beta, and if you're satisfied as well as a regular _latest_ (i.o.w. I won't wait until everything is done before releasing fixes).

sverweij:
B.t.w about _this_ issue: your expectation dependency-cruiser should work cross platform is correct. I wasn't aware of the limitations on windows (or of the differences between sh & zsh :eyes:) - so there's nothing to it than putting a layer of abstraction between it; `glob` (your suggestion) seems the perfect fit. 

Working on it :-)

sverweij:
@ajafff dependency-cruiser version 2.14.0-beta-0 with PR #36 (which is a fix for this issue) is on npmjs.org since a few minutes.

Let me know if it works for you.

sverweij:
Published 2.14.0 - which === 2.14.0-beta-1 except for the version number & npm tag - which closes this issue (on to PR reviewing then ...)

PR #36, opened 2018-02-16
Use 'glob' to abstract expansion mechanism across platforms

sverweij:
## Description
Should solve issue #32 

## Motivation and Context
dependency-cruiser should work across platform without modifications. 

## How Has This Been Tested?
- [x] regression test
- [x] unit tests on the added feature

## Types of changes
- ~~~[ ] Bug fix (non-breaking change which fixes an issue)~~~
- ~~~[ ] Breaking change (fix or feature that would cause existing functionality to change)~~~

## Checklist:

### DIFF
--- a/src/extract/gatherInitialSources.js
+++ b/src/extract/gatherInitialSources.js
@@ -1,5 +1,6 @@
 const fs            = require('fs');
 const path          = require('path');
+const glob          = require('glob');
 const ignore        = require('./ignore');
 const transpileMeta = require('./transpile/meta');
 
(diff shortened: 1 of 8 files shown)


src/extract/gatherInitialSources.js
```diff
@@ -41,14 +42,19 @@ function gatherScannableFilesFromDir (pDirName, pOptions) {
...
-                return pAll.concat(gatherScannableFilesFromDir(pThis, lOptions));
-            } else {
-                return pAll.concat(pThis);
-            }
-        },
-        []
-    );
+    return pFileDirArray
+        .reduce(
+            (pAll, pThis) => pAll.concat(glob.sync(pThis, {cwd:lOptions.baseDir})),
```
ajafff: performance tip: use the same options object for each call to glob and initialize the cache properties:

```js
const lGlobOptions = {
    cwd: lOptions.baseDir,
    cache: {},
    symlinks: {},
    statCache: {},
    realpathCache: {},
};
glob.sync(pFileDirArray[0], lGlobOptions);
glob.sync(pFileDirArray[1], lGlobOptions); // will use the cache of the previous call
```

sverweij: hmm, indeed. Didn't properly disable globbing in my manual tests :-/. Fix in 3804895 . I'm looking for a way to automate that test (at least locally).

Thanks for the performance optimisation tip; I'll look into it (measure the difference & if it's better put it in).

Issue #33, opened 2018-02-16
Add option to resolve symlinks of dependencies


ajafff:
### Expected Behavior
I use yarn workspaces to link the packages in my monorepo to each other. This bascially does `yarn link` (or `npm link`) behind the scenes.
The setup is as follows:

```
packages
  |-- a
  \-- b
node_modules
  |-- a -> ../packages/a
  \-- b -> ../packages/b
```

`a` requires `b`, so I would expect the path to be `packages/b/index.js` but instead it is `node_modules/b/index.js`.

Node resolves symlinks during require. That's why I expected this tool to do that too.

### Current Behavior
Symlinks are not resolved, see above.

### Possible Solution
Add an option whether `fs.realpath` should be used after resolving imports.
This should not be enabled by default, because it would probably break other users.

### Your Environment

* Version used: 2.13.1
* Node version: 9.5.0
* Operating System and version: macOS High Sierra 10.13.3
* Link to your project: https://github.com/fimbullinter/wotan (.dependecy-cruiser.json is not updated as of writing this)

sverweij:
Ouch! I was expecting that to work as well. Thanks for reporting. You're also right that fixing it right on will be breaking, so I'll put it behind an option (both in the cli and in .dependency-cruiser.json).

(Wotan looks cool b.t.w.! I'll definitely give it a closer look after this).

ajafff:
I thought this would be as easy as adding the `preserveSymlinks: false` option to `resolve.sync`, but it seems that resolving symlinks was never implemented in that module: https://github.com/browserify/resolve/issues/122

I'll probably try to fix this upstream.

ajafff:
I totally missed that. So it seems that's not a bug in `resolve`. They just don't use realpath on the result at all. Instead they resolve the input `basedir` if explicitly requested.

That means this package just needs a call to `fs.realpathSync` after using `resolve.sync`.

sverweij:
We still need the fix in dependency-cruiser, but even when yesterday I concurred with the view it isn't a bug in `resolve`, today I've had a closer look and I'd say it still needs a fix upstream as well - given that `resolve`'s intent is to implement the `require.resolve` algorithm:

Let's say we have a link and an index.js in a folder:
```
only-the-link@ -> the-real-dir
the-real-dir/
the-real-dir/index.js
```

Now I'd expect `resolve.sync` and node's `require.resolve` to have the same result. But they don't:
```javascript
> require.resolve('./only-the-link');
[...]

PR #37, opened 2018-02-17
Add option to preserve/resolve symlinks

ajafff:
## Description
Add option `preserveSymlinks` everywhere:
* resolve/index.js
  * works for both AMD and CJS
* extract.js
* validate.js
* defaults.json
* jsonschema (twice, don't know why)
* CLI
* docs
* tests

The option (currently) defaults to `true` which leaves the old behavior unchanged. This can and probably should be changed in a future major version to be consistent with Node's default behavior.
To disable this option, either specify it in the config file or use the CLI argument `--preserve-symlinks false`.

## Motivation and Context
Fixes: #33 

## How Has This Been Tested?
Tested in the unit tests. (needed to add a devDependency on `symlink-dir` for easier handling of symlinks on windows)
Also tested locally with in my project https://github.com/fimbullinter/wotan. Tested config file and CLI option.

## Types of changes

- ~~[ ] Bug fix (non-breaking change which fixes an issue)~~
- ~~[ ] Breaking change (fix or feature that would cause existing functionality to change)~~

## Checklist:

- [x] My code follows the code style of this project. (I hope so, this is not always obvious)

### DIFF
--- a/src/extract/extract.js
+++ b/src/extract/extract.js
@@ -24,15 +24,16 @@ const toTypescriptAST             = require('./parse/toTypescriptAST');
  *
  * @param  {string} pFileName path to the file
  * @param  {object} pOptions  an object with one or more of these properties:
- *                            - baseDir       - the directory to consider as the
- *                                              base for all files
- *                                              Default: the current working directory
- *                            - moduleSystems - an array of module systems to
- *                                              consider.
- *                                              Default: ["cjs", "es6", "amd"]
- *                            - exclude       - a regular expression string
- *                              with a pattern of modules to exclude (e.g.
- *                              "(node_modules)"). Default: none
+ *                            - baseDir         - the directory to consider as the
+ *                                                base for all files
+ *                                                Default: the current working directory
+ *                            - moduleSystems   - an array of module systems to
+ *                                                consider.
+ *                                                Default: ["cjs", "es6", "amd"]
+ *                            - exclude         - a regular expression string
+ *                                                with a pattern of modules to exclude
+ *                                                (e.g. "(node_modules)"). Default: none
+ *                            - preserveSymlink - don't resolve symlinks.
  * @return {array}           an array of dependency objects (see above)
  */
 module.exports = (pFileName, pOptions) => {
@@ -41,7 +42,8 @@ module.exports = (pFileName, pOptions) => {
             {
                 baseDir: process.cwd(),
                 moduleSystems: ["cjs", "es6", "amd"],
-                tsPreCompilationDeps: false
+                tsPreCompilationDeps: false,
+                preserveSymlinks: true
             },
             pOptions
         );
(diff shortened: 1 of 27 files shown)


sverweij:
Thanks! It looks very complete at a first glance. I'll try to review it more closely tomorrow.

sverweij:
I'll do the review in chunks so it is easier to follow & comment on

#### Your question about the jsonschema:
> jsonschema (twice, don't know why)
Good point;
- The _rule set_ has it in the options section for obvious reasons; 
- Less obvious is why it's in the (json) _output file_. That file has a section with the command line options used to generate it. I put it there for support, and the ability to put it in custom reporters. 
Maybe some day I should put the shared definitions in a separate schema and link to it from the other two schemas to be more DRY.

sverweij:
### command line & config option
#### switch vs option with a parameter
I see you didn't go for the 'switch', parameter-less of command line option, and I understand the dilemma:
- putting in a switch with the proper name (`--preserve-symlinks`, to mirror the one in node) would introduce a breaking change because leaving it out imply would imply the default for it would be `false`
- putting in a switch that doesn't introduce a breaking change (`--follow-symlinks`) could be confusing both in code and to people using the command line.

... so you did the only sensible thing left and opted for the proper name + config parameter in order to not introduce a breaking change. I appreciate that!

I've given it some thought and I'd rather do the breaking change right now (and 'semver-major' the release number) - which is consistent with your advice to stay aligned with node's default behaviour. 

#### Small things
I've taken the liberty to update the PR with some nitpicky things as well:
- the (line) test coverage fell below 100% because there was no test covering the command line validity check - this is partly my fault for not configuring a minimum coverage in the ci (fixed).
- The first and last options that appear when running depcruisie without parameters (/ or with --help) are the easiest to find. Which is why the `--init` option was on the last option (fixed).
- I've also added a thing so the followSymlinks appear in the `optionsUsed` in the extraction output object (& the json).

src/extract/extract.js
```diff
@@ -41,7 +42,8 @@ module.exports = (pFileName, pOptions) => {
             {
                 baseDir: process.cwd(),
                 moduleSystems: ["cjs", "es6", "amd"],
-                tsPreCompilationDeps: false
+                tsPreCompilationDeps: false,
+                preserveSymlinks: true
```
ajafff: this should be `false` to match the docs?

test/extract/fixtures/amd.json
```diff
@@ -247,5 +247,67 @@
...
+    },
+    {
+        "title": "require a symlinked module",
+        "input": {
+            "fileName": "test/extract/fixtures/amd/symlink.js"
+        },
+        "expected": [
+            {
+                "module": "../symlinked",
+                "resolved": "test/extract/fixtures/symlinked/index.js",
```
ajafff: this will change when the default is set to `false`

test/extract/fixtures/cjs.json
```diff
@@ -259,5 +259,89 @@
...
+    },
+    {
+        "title": "require a symlinked module",
+        "input": {
+            "fileName": "test/extract/fixtures/cjs/symlink.js"
+        },
+        "expected": [
+            {
+                "module": "../symlinked",
+                "resolved": "test/extract/fixtures/symlinked/index.js",
```
ajafff: same as above

PR #38, opened 2018-02-17
Find closest package.json

### LINKED ISSUE #35: Use closest package.json instead of ./package.json
### Expected Behavior

I have a monorepo with multiple packages. I want to cruise the whole repo at once to avoid unnecessary work.
To determine if a dependency of a file is an npm package declared in `package.json` depcruise looks at the package.json in the directory the command was executed.
In my case this is wrong, because each package has its own package.json and there is another one for development dependencies in the repo root.

```
package.json // dependencies: dependency-cruiser, glob
scripts
  \-- script.js // require('glob')
packages
  |-- a
  |   |-- index.js // require('foo')
  |   \-- package.json // dependencies: foo
  \-- b
      |-- src
      |    \-- index.js // require('bar')
      \-- package.json // dependencies: bar
```

Expected:
* `scripts/script.js` uses `./package.json` and is valid, because the dependency is listed
* `packages/a/index.js` uses `packages/a/package.json` and is therefore valid
* `packages/b/src/index.js` uses `packages/b/package.json` and is therefore valid

### Current Behavior
Only the root package.json is looked at. The dependencies declared in the package.json files in subdirectories are ignored.
`packages/a/index.js` and `packages/b/src/index.js` are invalid.

### Possible Solution
Always use the closest package.json for each file.

### Your Environment

* Version used: 2.13.1
* Node version: 9.5.0
* Operating System and version: macOS 10.13.3
* Link to your project: https://github.com/fimbullinter/wotan

ajafff:
## Description
Lookup the closest package.json for each cruised module. This is necessary to support monorepos that have multiple packages (with their own package.json, of course).

Related note: You might want to reconsider your decision not to cache the result.

As a side effect, if you follow node_modules (you probably shouldn't), transitive dependencies are validated with the package.json file of the package that uses it (as opposed to your package.json before the change). This is more correct than before but probably still not desired.

## Motivation and Context
Fixes #35

## How Has This Been Tested?
Unit test.
Tested in my monorepo https://github.com/fimbullinter/wotan. Before the change there were 144 violations. After the change only 6 violations remained, which are probably real violations.

## Types of changes
I'm not sure.
- ~~[ ] Breaking change (fix or feature that would cause existing functionality to change)~~

## Checklist:


sverweij:
This is a cool feature - and the changes needed for it look surprisingly straightforward. I'll give them a closer look tomorrow (🤞).

About the caching: I'll do some measurements again (on a slow machine with a spinning disk ...) to see if it now speeds things up.

ajafff:
I just upgraded to 3.0.0-beta-1 and it works great: https://github.com/fimbullinter/wotan/commit/3f5874305db47582dade1d774142ffd2624500bd

The changes from this PR and the symlink changes reduced the errors and warnings from over 100 to 6. These warnings were actually legitimate. After some config and dependency fixes I'm now down to zero violations.

sverweij:
cool!

I'm in the process of following your advice to look into caching - I've slapped a `_.memoize` around the package-json-looker-upper and it does seem to have a positive effect (~7% faster) on the two repos I've tried (babel and wotan :-)).

I was surprised how little impact your changes made to performance in the first place...

src/extract/resolve/readPackageDeps.js
```diff
@@ -46,3 +47,5 @@ module.exports =
         }
         return lRetval;
     };
+
+module.exports = _memoize(readPackageDeps);
```
ajafff: I guess this can be optimized further. Currently this only momoizes the result for each start directory. It doesn't cache the lookups further up the directory tree.
Consider the following structure:

```
package.json
dir-a
    |-- subdir-one
    |    |-- file-one.js
    |    \-- file-two.js
    \-- subdir-two
         \-- file-three.js
dir-b
    \-- file-four.js
```

The cache will be used once because `file-one.js` and `file-two.js` are in the same directory.
That means there are still 2 lookups for `dir-a` and 3 lookups for the current directory.

If `readPackageDeps` was recursively calling a memoized version of itself, you would read only once in each directory.

```js
const readPackageDeps =
    (pFileDir) => {
        let lRetval = null;

        try {
            // find the closest package.json from pFileDir
            const lPackageContent = fs.readFileSync(path.join(lLookupDir, 'package.json'), 'utf8');
            try {
                lRetVal = JSON.parse(lPackageContent);
            } catch (e) {
                // left empty on purpose
            }
        } catch (e) {
            const lNextDir = path.dirname(lLookupDir);

            if (lNextDir !== pFileDir) {
                // not yet reached root directory
                lRetval = momoizedReadPackageDeps(lNextDir);
            }
        }
        return lRetval;
    };

const memoizedReadPackageDeps = _memoize(readPackageDeps);

module.exports = memoizedReadPackageDeps;
```

ajafff: I can't really test this, because I only have SSDs and the performance difference between completely uncached and fully cached (code I posted above) is around 100ms from 2.35s to 2.25s.
The difference between your cache implementation vs mine is not really measurable.

sverweij: - this recursive implementation is more elegant than the `do ... while` - I like it! keeper :-)
- performance-wise it indeed seems a tie between my crude memoize-only and this one (even on humongous code bases) - which is a bit puzzling.
- your measurement of cached being ~4% faster than non-cached is consistent with my findings.

The relative differences between cached and uncached on an old MacBook with spinning disks were even smaller. I might do some more tests with my current Mac & external spinning disks to level out the differences in CPU, memory, ..., but not in the context of this PR - the differences are too small to worry about for now.

PR #42, opened 2018-02-27
fix:lack cruise ref

lin-hun:
the api doc lack the cruise ref;
detail on code


sverweij:
good catch @lin-hun, thanks for the PR!

PR #43, opened 2018-03-14
specify which path we're matching

jessitron:
This wasn't clear to me, when I run `depcruise --validate .dependency-cruiser.json src`, exactly what I was trying to match in that regex. (relative to here? relative to src? relative to the file, as passed to `import` ?)

Thank you for making this tool and publishing and documenting it! This was really useful yesterday in identifying a circular dependency in TypeScript, which mocha totally barfs on.

> do that.

## Types of changes

## Checklist:

- [ ] My code follows the code style of this project.

### DIFF
--- a/doc/rules-reference.md
+++ b/doc/rules-reference.md
@@ -137,7 +137,9 @@ out dependency-cruiser will assume it to be `warn`.
 A regular expression an end of a dependency should match to be catched by this
 rule.
 
-In `from`, this is the path from project root to the file containing a dependency. In `to`, this is the path from project root to the file the dependency resolves to.
+In `from`, this is the path from the current working directory (typically your 
+project root) to the file containing a dependency. In `to`, this is the path from
+the current working directory to the file the dependency resolves to.
 
 When path is in a `to` part of a rule it accepts the regular expression
 'group matching' special variables `$0`, `$1`, `$2`, ...  as well. See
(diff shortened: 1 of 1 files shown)


doc/rules-reference.md
```diff
@@ -137,6 +137,8 @@ out dependency-cruiser will assume it to be `warn`.
 A regular expression an end of a dependency should match to be catched by this
 rule.
 
+In `from`, this is the path from project root to the file containing a dependency. In `to`, this is the path from project root to the file the dependency resolves to.
```
ajafff: I wouldn't call this "project root". IMO "current working directory" matches better.

Something like: All paths are relative to the current working directory.

jessitron:
whatever is most accurate! the PR is really a suggestion for better docs
(this was the one thing that frustrated me using it yesterday, everything
else made sense and worked well).

sverweij:
@jessitron thanks for the PR :+1: ! It improves the doc where it was lacking. Sometimes it confuses me as well - but I never had the realization to document it.

doc/rules-reference.md
sverweij: The formulations supplement each other:
- technically dependency-cruiser matches from the current working directory
- typically you'll be working from the root of your project folder. Mentioning that will probably be helpful for first time users.

sverweij:
@jessitron thanks again for the PR - sorry it took so long to merge!

PR #62, opened 2018-09-09
bugfix(cli): if an exception occurs, return 1 as an exit code

sverweij:
## Description
Set the exit code to 1 if an exception occurs, instead of keeping it on 0.

## Motivation and Context
Currently, if an exception occurs (e.g. config file not found, can't write to output, error in the config file, ...) dependency-cruiser happily returns 0 as the exit code - it just outputs the error message on stderr. This means it's easier for those exceptions like that to remain undetected - e.g. running on your ci, or as part of a build script, both of which expect non-zero exit codes when something is wrong.

This PR corrects that, so that if something is amiss with running dependency-cruiser it's obvious immediately.

This PR also
- makes sure cli error messages get written to stderr in less spots
- refactors the process exiting so it's unit testable

## How Has This Been Tested?
- [x] additional unit tests
- [x] non regressing unit tests in the rest of the code

## Types of changes

## Checklist:

### DIFF
--- a/bin/dependency-cruise
+++ b/bin/dependency-cruise
@@ -11,7 +11,8 @@ if (!semver.satisfies(process.versions.node, $package.engines.node)) {
     process.stderr.write(`\nERROR: your node version (${process.versions.node}) is not recent enough.\n`);
     process.stderr.write(`       dependency-cruiser needs a version of node ${$package.engines.node}\n\n`);
 
-    process.exitCode = 1;
+    /* eslint no-process-exit: 0 */
+    process.exit(1);
 }
 
 program
@@ -45,12 +46,10 @@ program
     .parse(process.argv);
 
 if (Boolean(program.args[0]) || program.info || program.init) {
-    const lExitCode = processCLI(
+    process.exitCode = processCLI(
         program.args,
         program
     );
-
-    process.exitCode = lExitCode;
 } else {
     program.help();
 }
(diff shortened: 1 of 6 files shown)


src/cli/index.js
```diff
@@ -64,27 +66,27 @@ function runCruise(pFileDirArray, pOptions) {
...
             process.stdout.write(formatMetaInfo());
         } else if (pOptions.init === true){
             createRulesFile(pOptions);
         } else {
-            runCruise(pFileDirArray, pOptions);
+            lExitCode = runCruise(pFileDirArray, pOptions);
         }
     } catch (e) {
         process.stderr.write(`\n  ERROR: ${e.message}\n`);
+        lExitCode = -1;
```
ajafff: why `-1`? AFAICT (all common) shells use one byte unsigned integers as return codes. That means -1 is simply converted to 255.

bin/dependency-cruise
```diff
@@ -46,10 +46,12 @@ program
     .parse(process.argv);
 
 if (Boolean(program.args[0]) || program.info || program.init) {
-    processCLI(
+    const lExitCode = processCLI(
         program.args,
         program
     );
+
+    process.exit(lExitCode);
```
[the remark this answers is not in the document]
sverweij: Thanks! I'll change it.

src/cli/index.js
sverweij: No particular reason, just an old habit. It idd results in 255 on un*x/ posix platforms.

I'll see if there's something more appropriate - I'll probably settle for `1`. An unrepresentative sample from some of the tools in ts/js land: tsc, tslint, acorn use `1`, eslint uses `2`. (Suggestions welcome).

PR #89, opened 2019-01-01
feature(resolve): support monorepo 'combined' dependencies

sverweij:
## Description
In some monorepos the child packages rely on dependencies being declared/ pulled in by the parent package. There's various reasons to do so e.g. to make sure all child packages use the same version of a framework/ utilties/ ..., without having to update the child packages each time - and the child packages are never released independently.

Thanks @ferdiemmen for the fruitful discussion that led to this PR.

### todo
- [x] implement merge funciton
- [x] implement options switch
- [x] implement the additional traversal strategy

### current behaviour
dependency-cruiser will search 'up' from a module to the first package.json it encounters will use that for deriving the dependency-types (and whether or not a dependency is declared in a package.json).

### behaviour after the pull request is merged
- if no switch in a .dependency-cruiser config is added: same as the current behaviour
- if the `combinedDependencies` switch config `options` key is (1) present and (2) equals true:
  - dependency-cruiser will search 'up' from a module to the root of where the cruise started
  - it merges each package.json it encounters that way into a 'combined' package.json and will use that for deriving the dependency-types c.s.

On merging dependencies from the package.json 'closest' to the module wins from ones further up in the tree.

## Motivation and Context
There's at least one mono repo that uses this 'combined dependencies' pattern - so there will be more out there.

## How Has This Been Tested?
- [x] unit tests
- [x] automated non-regression
- [x] manual test on said mono repo & some other big'uns

## Types of changes

## Checklist:


src/cli/initConfig/config.js.template.hbs
```diff
@@ -177,8 +177,14 @@ module.exports = {
         /* if true detect dependencies that only exist before typescript-to-javascript compilation */
         // , tsPreCompilationDeps: false
 
-        /* if true leave symlinks untouched, otherwise use the realpath */
-        // , preserveSymlinks: false
+        /* if true detect dependencies that only exist before typescript-to-javascript compilation */
+        // , tsPreCompilationDeps: false
```
sverweij: tsPreCompilationDeps is duplicate now - doing away with the preserveSymlinks thing :stop_sign:

sverweij: fixed in 72da854

PR #119, opened 2019-03-27
bugfix(extract): also detect dynamic imports in ts-precompilation-deps

sverweij:
## Description Motivation and Context
See title. It doesn't and it should. Addition to #114 
Also accept only string literals in the CommonJS detection 

## How Has This Been Tested?
- [x] automated non-regression 
- [x] unit tests 

## Types of changes

## Checklist:


src/extract/ast-extractors/extract-typescript-deps.js
```diff
@@ -86,25 +86,39 @@ function isRequireCallExpression(pASTNode) {
         typescript.SyntaxKind[pASTNode.expression.originalKeywordKind] === 'RequireKeyword';
 }
 
+function isDynamicImportExpression(pASTNode) {
```
ajafff: There are also import types in case you want to handle them as well:

```ts
var v: import('./types').T;
```

They can also occur in JSDoc of JS files:

```js
/** @type {import('./types').T} */
var v;
```

You might be interested in my library `tsutils`. It has a function `findImports` that handles imports, import equals, re-exports, import expressions, import types and in JS files it also finds require and import types in JSDoc.

sverweij: Thanks! And idd - I'm using these type imports in jsdoc. I'll checkout tsutils. The less custom code the better.

(I also realised that I probably need to handle AMD as well to be complete :-/)

sverweij: I've taken a peek at tsutils - I'm going to use it, although in a separate PR (and import types, and JSDoc imports will get supported then automatically.

PR #142, opened 2019-06-08
feature(*): adds a 'reachable' attribute for dead wood detection

### LINKED ISSUE #140: "not-reachable" rule
### **Question (or feature request if it isn't possible at the moment)**

Is it possible to declare an "not-reachable" rule, which is an extension of the "no-orphans" rule, but also checks for modules with outgoing imports. This rule would require to specify a list of "entry points".

### Example

Given

```plain
src/index.js       -> src/mod_a.js
src/mod_a.js       -> src/mod_b.js
src/mod_orphan.js  -> src/mod_b.js
```

Then this rule should report `src/mod_orphan.js` as "not reachable" (given `["src/index.js"]` as "entry point list").

sverweij:
## Description
Adds ...
- a boolean `reachable` attribute you can use in the `to` part of rules. If you specificy this rule dependency-cruiser will check all modules matching the `path` and `pathNot` of the `to` path of the rule are _reachable_ from modules matched by the `from` part of the rule.

Sample of a rule:
```javascript
...
{
  forbidden: [
    {
      name: 'no-unreachable-from-api',
      from: {
        path: '^src/main/index.ts'
      },
      to: {
        path: '^src',
        pathNot: '\\.spec\\.ts$',
        reachable: false
      }
    }
  ]
}
...
```

- each module matched by the path & pathNot attributes in the to part of the rule will contain a `reachable` array that tells whether the module is reachable or not, and according to what rule.

Sample output snippet
```json
...
  {
    "source": "src/cli/utensils.ts",
    "reachable": [
      {
        "value": false,
        "asDefinedInRule": "no-unreachable-from-api"
      },
      {
        "value": true,
        "asDefinedInRule": "no-unreachable-from-cli"
      }
   ]
  }
...
```
> (In an earlier version of this PR the `reachable` attribute in the output was just a boolean - 
> this prohibited the existence of multiple reachability rules - which might be useful in e.g. 
> mono repos. 

In the _err_ reporter they show up as a regular 'module only' rule e.g.:
```
  warn no-unreachable-from-api: src/cli/utensils.ts

✖ 1 dependency violations (0 errors, 1 warnings). 201 modules cruised.
```

And the same for the  _dot_ reporter:
dependencies unreachable from src/index.js - reachable rule on

<details>
<summary>todo list</summary>

- [x] document
- [x] optimise performance? (current implementation is a straight BFS with no caching, running against the in-memory dependency-graph) => Implemented caching, which improved performance from 1.01 - 1.12x, depending a bit on the nature of the code base. Nowhere a worse performance though => keeper. 
- ~~handle the tech debt mentioned by code climate (dups)~~ -> we need to address this anyway to get in more matching stuff; future feature/ TD item
- [x] separate RuleType for this?
- [x] adapt the err reporter to produce sensible output. The standard `from -> to` reporting won't cut it (see #143 - useful for other module only rules too)
- ~~adapt the dot report(s?) to properly show this (i.e. showing anything with a false reachability with the same zombie-colour as orphans).~~ future feature
- ~~enable other attributes in rules with a `reachable` attribute, (enable group matching in the same)~~ => push to future feature

</details>

## Motivation and Context
fixes #140 

## How Has This Been Tested?
- [x] new automated unit/ integration tests
- [x] manually on some real life code bases

## Types of changes

## Checklist:


ulrichb:
Hi @sverweij, 
sorry for my late answer.

In general the new rule works great (detected 4 correct positive unused modules in my code base). 

I used the following (forbidden) rule: 

```js
{
    name: "no-unreachable-typescript-modules",
    severity: "error",
    from: {
        path: "^(Cli/index.ts|BackEnd/index.ts|FrontEnd/index.tsx)$"
    },
    to: {
        path: "\.(ts|tsx)$",
        pathNot: "(\.d\.ts$|^Tests/)",
        reachable: false,
    }
},
```

Reasoning: 
* The `\.(ts|tsx)$` is necessary to avoid false positives for all the (as self-referencing reported) JS-configuration modules (jest.config.js, ecosystem.config.js, .dependency-cruiser.js, ...).
  * This is okay for me because I'm only interested in TS modules anyways.
* The `.d.ts` exclusion is necessary for the (as self-referencing reported) declaration files.
* I had to completely exclude `^Tests/` because of test infrastructure modules next to my `.test.ts` modules.

**Regarding performance:**
* previous version: 7.2, 7.5, 7,4, 7.6 secs (four runs)
* 4.22.0-beta-2 (including the rule): 6.3, 4.8, 4.7, 4.5, 4.6 secs (five runs)

sverweij:
hi @ulrichb - thanks for your detailed reply - much appreciated!

I'll add the .d.ts to the example in the documentation.

From your measurements it seems the impact on performance is not too bad (_faster_ even on your code base - do I see that correctly?)

ulrichb:
> From your measurements it seems the impact on performance is not too bad (_faster_ even on your code base - do I see that correctly?)

Yes!

PR #200, opened 2019-10-09
Added types for ICruiseResult

tharun208:
> do that.

## Types of changes

## Checklist

- [ ] I have updated the documentation accordingly.

### DIFF
--- a/types/dependency-cruiser.d.ts
+++ b/types/dependency-cruiser.d.ts
@@ -406,13 +406,451 @@ export interface ICruiseOptions {
    */
   forceOrphanCheck?: boolean;
 }
+export interface ICruiseResult {
+  /**
+   * A list of modules, with for each module the modules it depends upon
+   */
+  modules: Module[];
+  /**
+   * Data summarizing the found dependencies
+   */
+  summary: Summary;
+}
+
+export interface Module {
+  /**
+   * Whether or not this is a node.js core module
+   */
+  coreModule?: boolean;
+  /**
+   * 'true' if dependency-cruiser could not resolve the module name in the source code to a
+   * file name or core module. 'false' in all other cases.
+   */
+  couldNotResolve?: boolean;
+  dependencies:     Dependency[];
+  /**
+   * the type of inclusion - local, core, unknown (= we honestly don't know), undetermined (=
+   * we didn't bother determining it) or one of the npm dependencies defined in a package.jsom
+   * ('npm' for 'depenencies', 'npm-dev', 'npm-optional', 'npm-peer', 'npm-no-pkg' for
+   * development, optional, peer dependencies and dependencies in node_modules but not in
+   * package.json respectively)
+   */
+  dependencyTypes?: DependencyType[];
+  /**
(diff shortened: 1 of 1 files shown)


types/dependency-cruiser.d.ts
```diff
@@ -441,7 +879,7 @@ export interface IReporterOutput {
  *                        [here](../src/cli/flattenTypeScriptConfig.js))
  */
 export function cruise(
-  pFileDirArray: string[],
+  pFileDirArray: ICruiseResult[],
```
sverweij: This should stay a `string[]` - it's the files 'n folders the consumer puts in. You might've confused this parameter with the one in the `format` function below :-)

sverweij:
Hi @tharun208 thanks for this PR! 

At a glance it's looking good - I'll be reviewing it on the short term. You might've seen a comment or two already when you read this. 

One thing I do have to ask though: could you tick the checkbox in the Checklist that says you've read (& understood) the [contributing ](https://github.com/sverweij/dependency-cruiser/blob/develop/.github/CONTRIBUTING.md) document? It a.o. ensures we can release your contribution under MIT license.

types/dependency-cruiser.d.ts
tharun208: Okay, I will get it fixed

types/dependency-cruiser.d.ts
```diff
@@ -406,13 +406,451 @@ export interface ICruiseOptions {
...
+}
+
+/**
+ * How severe a violation of a rule is. The 'error' severity will make some reporters return
+ * a non-zero exit code, so if you want e.g. a build to stop when there's a rule violated:
+ * use that. The absence of the 'ignore' severity here is by design; ignored rules don't
+ * show up in the output.
+ *
+ * Severity to use when a dependency is not in the 'allowed' set of rules. Defaults to 'warn'
+ */
```
sverweij: this piece of jsdoc pertains to the `severity` in `IRuleSummaryType` - can you move it up there?

tharun208: Sorry, Cant able to get this comment!

sverweij: I think the jsdoc belongs just over the `severity` attribute, like so:

```javascript
export interface IRuleSummaryType {
   /**
    * The (short, eslint style) name of the violated rule. Typically something like
    * 'no-core-punycode' or 'no-outside-deps'.
    */
   name: string;
   /**
    * How severe a violation of a rule is. The 'error' severity will make some reporters return
    * a non-zero exit code, so if you want e.g. a build to stop when there's a rule violated:
    * use that. The absence of the 'ignore' severity here is by design; ignored rules don't
    * show up in the output.
    *
    * Severity to use when a dependency is not in the 'allowed' set of rules. Defaults to 'warn'
    */
   severity: SeverityType;
 }
```

tharun208: sure got it

tharun208:
Hi @sverweij, I will give a fresh PR with all the issues fixed.

PR #216, opened 2019-11-16
feature(extract): enable detection of commonjs dependencies not (directly) declared with 'require'

### LINKED ISSUE #206: resolve window.require('path') dependencies
Electron.js app has a specific approach for using at compile-time (webpack bundling, etc.) and runtime dependencies. So for example, webpack bundles sources and don't handle 'window.require' functions. But at runtime 'window.require' works like in node.js 'require'

### Context
```ts
//file ./src/index.ts
import { Component } from '@angular/core'; // <-- this is dependent module at compile-time 
...
export class AboutNgComponent {
  private electron: any;
  public get version(): string {
    return this.electron.remote.app.getVersion();
  }
  construnctor() {
    this.electron = window.require('electron');  // <-- this is dependent module at runtime
  }
}

```

### Expected Behavior
handle 'window.require' dependencies

### Current Behavior
doesn't find 'window.require' modules

### Possible Solution
I can't use 'require' instead of 'window.require' - see. https://github.com/electron/electron/issues/7300
So I hope, there are ability to add some extensions to handle specific cases like current

sverweij:
## Description

- Adds an `exoticRequireStrings` option to the dependency-cruiser config, that takes zero or more alternative strings dependency-cruiser should treat like a function that requires other modules (see the added rules documentation for details).
  - Added (commented) to the init templates
  - Document in the rules reference and the FAQ
  - Works for all supported languages
- Currently limited to single functions (`need('exodash')`, not yet nested ones (`window.require('elektrodash'`)

## Todo

- [/] make it possible to define rules on exotic require strings (either forbidding or allowing things based on the require string used) => separate PR
- [/] implement in the typescript extraction for composites ('window.require') => separate PR
- [/] implement in the javscript extraction for composites ('window.require') => separate PR

## Motivation and Context

- first step on the way to address #206
- also addresses some similar use cases (hopefully in the decline) where it's necessary to override or wrap the `require` function but where you still want or need to infer dependencies.

## How Has This Been Tested?

- [x] new unit tests

## Types of changes

## Checklist

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -262,6 +262,14 @@ _externalModuleResolutionStrategy_ key:
 emits the circular path in the _err_, _err-long_, _err-html_ and
 _teamcity_ reporters.
 
+### Q: I'm using window.require or a require wrapper - how do I make sure dependencies I declared like that are included?
+
+**A**: Add a _exoticRequireStrings_ key in your configuration with the 
+
+```json
+"exoticRequireStrings": ["window.require"]
+```
+
 ## Expanding dependency-cruiser
 
 ### Q: How do I add a new output format?
--- a/doc/rules-reference.md
+++ b/doc/rules-reference.md
@@ -1091,12 +1091,34 @@ the file it investigates:
 
 - monodash/
 - **package.json** _<- look in this one as well; merge it into the one down the tree_
-- packages/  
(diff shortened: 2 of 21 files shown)


trongham (commented):
Nice

PR #243, opened 2020-01-14
add .d.ts to allow for custom tsconfig resolving

davidparkagoda:
In one of our projects we use `.d.ts` files which are not generated to limit the possibility in the file. Most of them are model files

> do that.

## Description
Just add `.d.ts` the default is to only have `.ts` and `.tsx`. Unsure if this is the correct place though.

## Motivation and Context
In one of our projects we use `.d.ts` files which are not generated to limit the possibility in the file. Most of them are model files

## How Has This Been Tested?
Tested locally on the project

## Types of changes

## Checklist

- [ ] The code I add will be subject to [The MIT license](../LICENSE), and I'm OK with that.
- [ ] The code I've added is my own original work.
- [ ] My code follows the code style of this project.
- [ ] I have updated the documentation accordingly.

### DIFF
--- a/src/main/resolveOptions/normalize.js
+++ b/src/main/resolveOptions/normalize.js
@@ -61,7 +61,17 @@ function compileResolveOptions(pResolveOptions, pTSConfig) {
   if (pResolveOptions.tsConfig && _has(pTSConfig, "options.baseUrl")) {
     lResolveOptions.plugins = pushPlugin(
       lResolveOptions.plugins,
-      new TsConfigPathsPlugin({ configFile: pResolveOptions.tsConfig, extensions: [".ts", ".tsx", ".d.ts"] })
+      new TsConfigPathsPlugin({
+        configFile: pResolveOptions.tsConfig,
+        // different from the typescript compiler, tsConfigPathsPlugin
+        // only has .ts and .tsx as default extension to scan given
+        // an import/ require/ export argument.
+        // Added .js, .jsx and .d.ts before .ts and .tsx to mirror
+        // the typescript compiler's behavior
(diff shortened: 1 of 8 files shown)


sverweij:
hi @davidparkagoda thanks for not only spotting the issue but creating a PR for it as well 😍! 

It took me a little while to figure out what issue it solved, but it looks like that when both _tsconfig aliasses/ paths_ and files with a _.d.ts_ extension are used the tsconfig-paths-webpack-plugin skips them (check?). Which is unexpected as it differs from the typescript compiler behavior given the same tsconfig.

If that was indeed the issue, the spot you've chosen to fix it is entirely correct. I'll have a closer look tonight/ later this week and merge release a.s.a.p. (at least in the beta channel). 

My hunch is that in addition to this PR the tsconfig-paths-webpack-plugin should be mended as well as (as its default differs from what tsc uses). And I wonder how it handles files like `dostuff.service.ts`, which have the same filename pattern as `types.d.ts` ...

davidparkagoda:
Yea sorry I did not clarify, hard to remove assumptions coming from my side.

To clarify, inside our project we have `.d.ts` files to make sure typescript does not generate javascript.

`Model.d.ts`
```
export interface Model {
   property: string
}
```

I needed this amendment for dependency cruiser to see the file and resolve the contents inside.

sverweij:
It seems there's more amiss with tsconfig-paths-webpack-plugin _extensions_ default - at least if we want it to match behave the same as the typescript compiler. tsc will look at js, json and jsx before it takes d.ts, ts and tsx into account (in that order), so I've added that to the PR.

Before merging the PR I'll add some automated tests to prevent regression.

PR #309, opened 2020-06-03
fix "typo" in Readme about config file extension

hoschi:
Minor fix :smiley:


sverweij:
Well spotted, thanks @hoschi - must have missed that one when switching the default format from json to javascript...

PR #329, opened 2020-07-05
feature(validate): enables using 'required' rules

### LINKED ISSUE #328: Feature request: Check if any import matches
### Context

Our codebase is moving to a new arch and I want to add a warning if a module hasn't been migrated yet. Essentially, I want to check if a type of dependency exists.

Let's say I want to check if all the controller modules have been migrated, which I can simplify into a check to see if they extend the new base controller, for eg: 

`class 'AbcController' extends 'GenericController'`

### Expected Behavior

I want to check if AbcController contains an import for GenericController:

```
    {
      name: 'no-clean-arch-controller',
      severity: 'warn',
      "from": { "path": "^src\/api\/([^/]+)\/([a-zA-Z]+Controller.ts)"},
      "to": { 
        "pathNot": "src\/framework\/GenericController.ts" 
      },
    }
```
with the result:

```
no-clean-arch-controller: src/api/x/abcController.ts → src/framework/GenericController.ts
```

### Current Behavior

It currently tells me about all the imports that aren't GenericControllerImpl, and nothing

```
no-clean-arch-controller: src/api/x/abcController.ts → src/example/some-import.ts
no-clean-arch-controller: src/api/x/abcController.ts → src/example/another-import.ts
```

### Possible Solution

A boolean to signify that there's no dependency from A to B would be helpful.

### Considered alternatives

I tried to make it work with _allowed_, but it became too complex and didn't solve my issue anyway.

sverweij:
## Description

Enables the specification of 'required' rules; rules that put restrictions on what modules must at least depend on. For example (adapted from @hzaun's description in #328)

```javascript
{
  required: [{
    name: "controllers-inherit-from-base",
    comment: "Each controller should inherit from the framework's base controller",
    module: {
      // every module that matches the pattern specified in path & pathNot ...
      path: "-controller\\.js$",
      pathNot: "^src/framework/base-controller\\.js$",
    },
    to: {
      // ... must depend at least once on the framework's base controller
      path: "^src/framework/base-controller\\.js$",
    }
  }]
}
```

It supports

- `path` and `pathNot` criteria for selecting modules to apply the _required_ rule to.
- `path` criterium (including [group matching](https://github.com/sverweij/dependency-cruiser/blob/develop/doc/rules-reference.md#group-matching)) for describing which dependencies must be matched.

This idea can be expanded upon with other attributes (_reachable_ jumps to mind). This PR, however, concentrates on the need behind issue #328. 

List of changes:
- [x] enable specifying required rules and make them part of the validation step (including updating the json schema)
- [x] normalise the required rules before use
- [x] enable use of group placeholders in the dependency-criteria of required rules
- [x] update the  configuration merge algorithm to take required rules into account
- [x] add required rules the teamcity reporter's output (in all other reporters this comes out of the box)
- [x] update typescript types to match the changes in the json schema
- [-] ~~enable reachable?~~ future feature

## Motivation and Context

fixes #328 

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/src/validate/is-module-only-rule.js
+++ b/src/validate/is-module-only-rule.js
@@ -4,9 +4,10 @@
  */
 function isModuleOnlyRule(pRule) {
   return (
-    Object.prototype.hasOwnProperty.call(pRule.from, "orphan") ||
+    Object.prototype.hasOwnProperty.call(pRule.from || {}, "orphan") ||
+    // note: the to might become optional for required rules
     Object.prototype.hasOwnProperty.call(pRule.to, "reachable") ||
-    Object.prototype.hasOwnProperty.call(pRule.to, "module")
+    Object.prototype.hasOwnProperty.call(pRule, "module")
   );
 }
 
--- a/src/validate/is-module-only-rule.js
+++ b/src/validate/is-module-only-rule.js
@@ -5,7 +5,8 @@
 function isModuleOnlyRule(pRule) {
   return (
     Object.prototype.hasOwnProperty.call(pRule.from, "orphan") ||
-    Object.prototype.hasOwnProperty.call(pRule.to, "reachable")
+    Object.prototype.hasOwnProperty.call(pRule.to, "reachable") ||
+    Object.prototype.hasOwnProperty.call(pRule.to, "module")
   );
 }
 
--- a/tools/schema/rule-set.mjs
+++ b/tools/schema/rule-set.mjs
@@ -33,7 +33,12 @@ const RULE_SET_TYPE_PROPERTIES = {
     },
     required: {
       type: "array",
-      despcription: "",
+      despcription:
+        "A list of rules that describe what dependencies modules _must_ have. " +
+        "E.g." +
+        " - every controller needs to (directly) depend on a base controller." +
+        " - each source file should be the dependency of a spec file with the same " +
+        "   base name",
       items: {
         $ref: "#/definitions/RequiredRuleType",
       },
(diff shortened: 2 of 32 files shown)


tools/schema/rule-set.mjs
```diff
@@ -31,6 +31,13 @@ const RULE_SET_TYPE_PROPERTIES = {
         "Severity to use when a dependency is not in the 'allowed' set of rules. " +
         "Defaults to 'warn'",
     },
+    required: {
+      type: "array",
+      despcription: "",
```
sverweij, line 36: Goed verhaal. Lekker kort.

src/validate/is-module-only-rule.js
```diff
@@ -5,7 +5,8 @@
 function isModuleOnlyRule(pRule) {
   return (
     Object.prototype.hasOwnProperty.call(pRule.from, "orphan") ||
-    Object.prototype.hasOwnProperty.call(pRule.to, "reachable")
+    Object.prototype.hasOwnProperty.call(pRule.to, "reachable") ||
+    Object.prototype.hasOwnProperty.call(pRule.to, "module")
```
sverweij, line 9: How come this works? Did you mean ...
```javascript
Object.prototype.hasOwnProperty.call(pRule, "module")
```

sverweij: :-o that fails idd. Not in the last place because it wasn't used at all in the required rule check => fixed

- [ ] slap an extra automated test or two on that

sverweij: fixed in 2351a0b

tools/schema/rule-set.mjs
sverweij: fixed

PR #342, opened 2020-08-05
ci(gh): try gh's new codeql-analysis thing

sverweij:
## Description

See title. It's in beta and might be interesting => try it in a PR. 
(It does a bit like LGTM which I tried earlier when GH acquired them - which was kicked out of dc because it was slow and didn't find anything other tools didn't, so let's see ...)

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,54 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [develop, master]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [develop]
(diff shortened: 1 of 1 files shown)


sverweij:
didn't find any new reasons to start using this. 

Also: scans _everything_ - also code that is deliberately wrong (e.g. mocks in the _test_ tree), which it shouldn't. Even though it scans these test mocks, it doesn't find anything wrong. So either it's _incredibly_ smart and context-aware (but why not skip it then?) or it's not yet there. 

Close without merge.

Issue #343, opened 2020-08-05
Feature request: Filter based on React (or other) Component imports


forivall:
### Context

I'm trying to use static analysis to determine the possible component tree within a large application, but omit all of the util / data / etc. cruft.

### Expected Behavior

Example:

`enum.ts`
```ts
export enum MyEnum {
  A = 1;
  B = 2;
```

`a.tsx`

```ts
import * as React from 'react'
import { MyEnum } from './enum'
import B from './b'
export default function A = (props) => <div key={MyEnum.A}><B /></div>
```

`b.tsx`
```ts
import * as React from 'react'
import { MyEnum } from './enum'
export default function B = (props) => <span>This is B = {MyEnum.B}</span>
```

with this option, the graph would be like
```dot
digraph {
  "a.ts" -> "b.ts"
}
```
and ignore the enum import

An even better use case would be to trace the actual component name, but that's a lot harder to decipher.

### Current Behavior

This feature doesn't exist, although it's similar to `tsPreCompilationDeps`

### Possible Solution

Analyse the code, and only count it as a dep if the import is used as a component in JSX. (This wouldn't cover all component uses, but it's most of them)

### Considered alternatives

Alternative is to use storybook, and then use react devtools to look at the tree, but that takes a lot of setup, and the thing i'm looking at is stuff we want to remove.

Other alternative i just thought of is to include only component files -- ie, only .tsx files. i'll be doing that instead for now.

forivall:
yup, my alternative to `--include-only` just tsx worked for my needs. feel free to close if this isn't something you'd pursue

forivall:
Or actually, another implementation possiblity is to filter only files that depend on `react` -- there's a couple `.tsx` files in my project that don't actually import `react`. -- better solution to this problem -- add an option _like_ `--focus`, that allows external ... looking at config typings

yeah, adding a new option to IIncludeOnlyType like `jsx: boolean` or `requires: string | string[]` would do the trick.

forivall:
Actually, i'd probably still want to do this, but spin it out to a separate module, and pull in dependency-cruiser's rendering internals. My solution in #344 wasn't perfect, because it still runs afoul of the index.tsx issue, which just re-exports other dependencies, which lead to this mess (rendered using xdot, as it was the only piece of existing software that highlights the edges on hover -- something else i'd want to make, a better visualized dot output. alas.)

[...]

PR #346, opened 2020-08-06
feat(wrap-stream-in-html): highlight connected edges

forivall:
## Description

Highlights connected edges to hovered node, and connected nodes to hovered edges.

## Motivation and Context

https://github.com/sverweij/dependency-cruiser/issues/343#issuecomment-670085693 you said it's on your wishlist, and i knew i could bang it out quickly. 😄 

## How Has This Been Tested?

I just launched it in the browser and hacked away. I was feeling nostalgic so it's es5.

## Screenshots

Screen Shot 2020-08-06 at 12 58 06 PM

Screen Shot 2020-08-06 at 12 58 13 PM

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


forivall:
When i update the docs, i'd also like to mention this bookmarklet i just made, so that clicking opens vscode instead of in the browser https://gist.github.com/forivall/373425de70964c8c271020cab3650048
```js
document.querySelector('svg').querySelectorAll('a').forEach(function (a) {
  var u = new URL(a.href.baseVal, document.location.href);
  u.protocol = 'vscode:';
  u.pathname = u.pathname.replace(/^\/*/, '//file/');
  a.href.baseVal = u.href;
});
```

sverweij:
@forivall whoa - that looks nice!! I'll give it a spin tomorrow!

About the bookmarklet - that's creative! Did you consider using the [prefix](https://github.com/sverweij/dependency-cruiser/blob/develop/doc/rules-reference.md#prefix-prefix-links-in-reports) option? I didn't know about the `vscode` protocol, but it totally works:

```javascript
  // in the options section of .dependency-cruiser.js plonk something like this:
  prefix: `vscode://file/${process.cwd()}/`,
```

```sh
# or in a command line variant for it:
depcruise -v -T dot --prefix "vscode://file/`pwd`/" -- src | dot -T svg | depcruise-wrap-stream-in-html | browser
```

That way the links in the dot (/svg) get this as a prefix. The documentation currently has an example that prefixes it with the base url of an online repository, but it's usable with other protocols as well.

forivall:
Yeah, I just saw the prefix option earlier. Once again, you've already got something there :P

sverweij:
@forivall thanks for this! It makes a huge difference in the usability of these reports. Nice touch to also highlight both nodes at the end of an edge on hover - and a good call to leave the existing css in place so there's a fallback for folks who disabled their javascript.

sverweij:
(the results look too pretty not to have in the cli documentation - so I've added :-) )

PR #354, opened 2020-08-25
feature(depcruise-fmt): add filter options

sverweij:
## Description

- Adds three command line filtering options to depcruise-fmt (--include-only, --focus, --exclude) that work the same as the ones on the depcruise command.
- Also adds short versions for the filter commands to the `depcruise` command (-I and -F respectively) to be consistent with depcruise-fmt and to make it easier to use if you're using the cli to interrogate.
- On the fly fixes an as of yet undetected bug where a reportOptions.dot/archi/ddot.filters.focus wasn't taken into account.

## Motivation and Context

`depcruise-fmt` is useful if you want to display the results of the same cruise in different ways, without having to run the cruise repeatedly. Especially on bigger code bases this can save time. Cruising all code can sometimes take more than a minute, while formatting usually takes well below a second.

Adding filter options increases the usefulness of the `depcruise-fmt` command. 

Cruise your humongous code base once (~30 - 60s?) and save it to a json (e.g. results.json):
- run depcruise-fmt several times against result.json to produce overviews of parts of the code base
- run depcruise-fmt with `--focus` to quickly find everything going into and coming out of packages/old-utils/spelunk-this.js
- run depcruise-fmt without any filters to find all violations
- run depcruise-fmt with submodules as filter to find violations for only that submodule

depcruise-fmt typically takes ~1s for each of this runs, so if you have a lot of things like this going on it could be big win™️. 

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/src/main/options/normalize.js
+++ b/src/main/options/normalize.js
@@ -1,3 +1,4 @@
+/* eslint-disable security/detect-object-injection */
 const normalizeREProperties = require("../utl/normalize-re-properties");
 const defaults = require("./defaults.json");
 
@@ -24,7 +25,7 @@ function normalizeReporterOptions(pReporterOptions) {
     "archi.filters.exclude.path",
     "dot.collapsePattern",
     "dot.filters.includeOnly.path",
-    "dot.filters.focus",
+    "dot.filters.focus.path",
     "dot.filters.exclude.path",
     "ddot.collapsePattern",
     "ddot.filters.includeOnly.path",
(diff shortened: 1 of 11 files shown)


src/main/options/normalize.js
```diff
@@ -24,7 +25,7 @@ function normalizeReporterOptions(pReporterOptions) {
     "archi.filters.exclude.path",
     "dot.collapsePattern",
     "dot.filters.includeOnly.path",
-    "dot.filters.focus",
+    "dot.filters.focus.path",
```
sverweij, line 28: (this likely fixes an as-of-yet undetected bug)

PR #393, opened 2020-11-13
support Svelte

truongsinh:
## Motivation and Context

Add support for [https://svelte.dev](https://svelte.dev). It's quite trending that even [Mozilla has a tutorial](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Client-side_JavaScript_frameworks/Svelte_getting_started).

## How Has This Been Tested?

Unit test and manually test against my repo.

## Screenshots

graphviz

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


truongsinh:
I have no idea how those unit tests are failing on CI

sverweij:
@truongsinh thanks for this PR - it looks like a great addition to dependency-cruiser's abilities. I'll review the PR on the weekend.

> I have no idea how those unit tests are failing on CI

- It seems the failing unit tests assume svelte2tsx is available. On the ci it isn't at the moment - likely because it's not in the devDependencies
- The linter step (only executed on node 14 a.t.m.)  complains about formatting - running `yarn lint:fix` (or `npm run lint:fix`) will typically fix that. (I'd have expected husky/ lint-staged would've taken care of it on commit already).

doc/faq.md
```diff
@@ -215,6 +215,14 @@ For `.vue` single file components it uses the `vue-template-compiler`
 
 - which will be in your module dependencies if you're developing with Vue).
 
+### Q: Does this work with Svelte as well?
+
+**A**: Yes.
+
+For `.svelte` single file components it uses the `svelte2tsx`
+
+- you might have to install it as peer dependency `npm i -D svelte2tsx`.
```
sverweij, line 224:
```suggestion
- you might have to install it as development dependency if you don't have it already: `npm i -D svelte2tsx`.
```

doc/faq.md
```diff
@@ -215,6 +215,14 @@ For `.vue` single file components it uses the `vue-template-compiler`
 
 - which will be in your module dependencies if you're developing with Vue).
 
+### Q: Does this work with Svelte as well?
+
+**A**: Yes.
+
+For `.svelte` single file components it uses the `svelte2tsx`
```
sverweij, line 222: During the ci-cleanup I noticed it also needs the svelte compiler, so `svelte` would be required to be present in the same node_modules tree as dependency-cruiser as well.

But I guess that's kind of a given when you're working on a svelte project (check)?

sverweij (commented):
As said earlier - thanks for this PR, @truongsinh - it's a nice addition to what dc can do. I do have a few questions on some of the implementation choices, so I understand why the code is as it is (see below).

truongsinh:
I push a totally new commit, but it seems the linter is really outdated and do not understand null-safe operator
`...pTranspileOptions?.tsConfig,`

package.json
```diff
@@ -210,15 +212,16 @@
     "types/**"
   ],
   "engines": {
-    "node": "^10||^12||>=13"
+    "node": "^10.17||^12||>=13"
```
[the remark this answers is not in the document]
truongsinh: Originally, I thought it was for `fs/promises`, but it turns out 10.0 already supports. Reverted.

src/extract/transpile/svelte-wrap.js
```diff
@@ -0,0 +1,20 @@
+const tryRequire = require("semver-try-require");
+const svelte2tsx = tryRequire(
+  "svelte2tsx",
```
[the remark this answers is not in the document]
truongsinh: Actually, I had a conversation in the Svelte group and they also suggested the same. New code pushed.

src/report/dot/default-theme.json
```diff
@@ -91,6 +91,10 @@
       "criteria": { "source": "\\.tsx$" },
       "attributes": { "fillcolor": "#bbfeff" }
     },
+    {
+      "criteria": { "source": "\\.svelte$" },
+      "attributes": { "fillcolor": "#febbff" }
+    },
```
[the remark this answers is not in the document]
truongsinh: Thanks 😉

test/extract/transpile/svelte-wrap.spec.js
```diff
@@ -0,0 +1,26 @@
...
+);
+
+describe("svelte transpiler", () => {
+  it("tells the svelte transpiler is available", () => {
+    expect(wrap.isAvailable()).to.equal(true);
+  });
+
+  it("'transpiles' svelte", async () => {
+    const observedPromise = fsPromise
+      .readFile("./test/extract/transpile/fixtures/svelte.svelte", "utf8")
```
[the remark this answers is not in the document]
truongsinh: Well, it's the same regardless of whether it's browser or server, https://nodejs.org/en/docs/guides/blocking-vs-non-blocking/ . It's just faster, and as I understood, this nodejs process also yields CPU resources to other processes when it's doing async IO task and no microtask left in eventloop (while with sync IO task, it will hold on to CPU resources, but do not utilize it)

doc/faq.md
truongsinh: no longer relevant due to https://github.com/sverweij/dependency-cruiser/pull/393#discussion_r523383294

doc/faq.md
truongsinh: Yep, https://github.com/sverweij/dependency-cruiser/pull/393#discussion_r523383294

sverweij:
> I push a totally new commit, but it seems the linter is really outdated and do not understand null-safe operator
> `...pTranspileOptions?.tsConfig,`

Optional chaining doesn't exist in node 10 and 12, so the node runtimes on the ci complain. As long as node 10 and 12  are on LTS, dependency-cruiser has to be able to run on them. The dependency-cruiser codebase currently uses `lodash/get`, which does support optional chaining, albeit with a different syntax: `...get(pTranspileOptions, 'tsConfig')`

truongsinh:
> > I push a totally new commit, but it seems the linter is really outdated and do not understand null-safe operator
> > `...pTranspileOptions?.tsConfig,`
> 
> Optional chaining doesn't exist in node 10 and 12, so the node runtimes on the ci complain. As long as node 10 and 12 are on LTS, dependency-cruiser has to be able to run on them. The dependency-cruiser codebase currently uses `lodash/get`, which does support optional chaining, albeit with a different syntax: `...get(pTranspileOptions, 'tsConfig')`

Oh yeah sorry I thought I was coding on TypeScript

test/extract/transpile/svelte-wrap.spec.js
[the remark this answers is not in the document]
sverweij: No new reasons I wasn't aware of :-) It's fine to keep using async code, but I'm always curious for new reasons to actually start using it in a command line tool.

It seems the nice and elegant fs.promises is not supported on node 10 until 10.17, though and I'd rather not major-bump over functionality that is only used in unit tests...

> The performance advantage is there for concurrent programs, where other javascript can be executed while
> during the execution of i/o. In these cases async is actually the only way to have decent functionality.

src/extract/transpile/svelte-wrap.js
```diff
@@ -0,0 +1,46 @@
...
+          if (attributes.lang !== "ts" || !pTsWrapper.isAvailable()) {
+            return { code: content };
+          }
+          const compiledToTsFromJs = pTsWrapper.transpile(content, {
+            ...pTranspileOptions,
+            tsConfig: {
+              ..._get(pTranspileOptions, "tsConfig", {}),
+              options: {
+                ..._get(pTranspileOptions, "tsConfig.options", {}),
+                importsNotUsedAsValues: "error",
```
sverweij, line 28: 1. What is the reason `importsNotUsedAsValues` is on `error` instead of one of the other values (e.g. `preserve` (or the default `remove`)? If I see correctly with this value on `error` the typescript compiler will actually throw, while the imports are still perfectly analysable, which dependency-cruiser's goal here.
2. Is there a reason to have `importsNotUsedAsValues` and `jsx` prevail over their counterparts set in the tsConfig?

test/extract/transpile/svelte-wrap.spec.js
```diff
@@ -0,0 +1,26 @@
...
+      .then((pSourceCode) => wrap.transpile(pSourceCode));
```
sverweij, line 15: The `expect` will likely fail on windows (because of line-end shenanigans). To prevent that other wrapper unit tests use `normalise-newline` on both the output of the read file and that of the wrap.transpile before passing it to the expect.

src/extract/transpile/svelte-wrap.js
```diff
@@ -0,0 +1,37 @@
+const tryRequire = require("semver-try-require");
+
+const svelte = tryRequire(
+  "svelte",
+  require("../../../package.json").supportedTranspilers.svelte
+);
+
+// eslint-disable-next-line import/order, node/no-unpublished-require
+const svelteCompile = svelte ? require('svelte/compiler') : false;
```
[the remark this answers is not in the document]
truongsinh: I did try that before, but `svelte.compiler` is undefined. You can check it by changing this line and run `npx mocha test test/extract/transpile/svelte-wrap.spec.js

src/extract/transpile/svelte-wrap.js
truongsinh: 1. The default `remove` will definitely not work, because of the way we use `svelte-preprocessor` and `svelte/compile`. If you try this value and run `npx mocha test test/extract/transpile/svelte-wrap.spec.js`, the test will fail (missing `import "./Header.svelte";`). I did try with `preserve`, and it seems to be no different than `error`, so I have no strong opinion on that, but the reason I originally chose `error` is because I got it from https://www.npmjs.com/package/@tsconfig/svelte

2. `importsNotUsedAsValues` is explain aboved. As for `jsx`, it's because it's assumed that in Svelte, it's "JS in HTML", not "HTML in JS", and thus the code inside `<script>` tag will be valid JS/TS only, not with JSX extension.

test/extract/transpile/svelte-wrap.spec.js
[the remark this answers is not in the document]
truongsinh: Turn our, `fs.promise` is already on 10 https://nodejs.org/en/blog/release/v10.0.0/  https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_fspromises_readfile_path_options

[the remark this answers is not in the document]
sverweij: You're right! Wonder where that linter warning about 10.17 came from

src/extract/transpile/svelte-wrap.js
sverweij: 1. Thanks for the explanation! Between `error` and `preserve`, with dependency-cruiser's use in mind `preserve` seems like a better fit; `error` will make the typescript compiler throw with an error that isn't relevant for dependency analysis.

2. Clear!

src/extract/transpile/svelte-wrap.js
```diff
@@ -0,0 +1,46 @@
...
+          });
+          return { code: compiledToTsFromJs };
+        },
+      }
+    );
+    const compiledSvelteCode = svelteCompile.compile(
+      optionallyCompileTsToJsInSvelte.code
+    );
+    // remove `import {...} from "svelte/internal";`
+    return compiledSvelteCode.js.code.replace(
```
sverweij, line 41: This code removes a real dependency from the output. I think it'd be better to return the `.js.code` as-is. If the end user doesn't want to see dependencies to `svelte/internal` she can always exclude them herself (either in config or in a command line switch).

src/extract/transpile/svelte-wrap.js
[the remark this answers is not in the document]
sverweij: Clear!  (And passing 'svelte/compiler' to semver-try-require doesn't work either as that doesn't support requiring submodules yet)

sverweij:
b.t.w. @truongsinh I've added an exception in dependency-cruiser's own rule config, so we can use `require('svelte/compiler')` without complaints and the ci can pass on to linting and testing steps.

B.t.w. linting is fine now, and all tests pass - but it seems not all test paths were covered yet. It's easy to check this locally: run `npm run test:cover` and open the report in `coverage/index.html` to see which paths need to be covered still.

src/extract/transpile/svelte-wrap.js
truongsinh: Ok I will use `preserve`

src/extract/transpile/svelte-wrap.js
truongsinh: ack

truongsinh:
address feedback:
- use importsNotUsedAsValues: "preserve" insteaf of error
- keep `import {...} from "svelte/internal";` and document this for users to decide to ignore it
- increase diff-test-coverage to 100%

btw, I use the recommended pattern https://mochajs.org/#dynamically-generating-tests

sverweij:
Looking good @truongsinh, thanks!

- In the mean time semver-try-require has had a small update so it can handle requiring 'svelte/compiler' directly - so I'll push a commit to this branch that uses that.
- And I'm a little worried about the async'ness of the svelte-wrapper's `transpile`. I don't see a way around it, but it's now returning a Promise (that if resolves to a string if all goes well) - so to-javascript-ast.js should handle that async'ness as well, but it doesn't (or I'm misunderstanding something fundamental on async/await and promises ...)

truongsinh:
> Looking good @truongsinh, thanks!
> 
> * In the mean time semver-try-require has had a small update so it can handle requiring 'svelte/compiler' directly - so I'll push a commit to this branch that uses that.
> * And I'm a little worried about the async'ness of the svelte-wrapper's `transpile`. I don't see a way around it, but it's now returning a Promise (that if resolves to a string if all goes well) - so to-javascript-ast.js should handle that async'ness as well, but it doesn't (or I'm misunderstanding something fundamental on async/await and promises ...), which means we might run into race conditions sooner or later.

you mean the current API for a transpiler is
```js
{
  isAvailable: () => bool
  transpile: () => string
}
```

and NOT
```

```js
{
  isAvailable: () => bool
  transpile: () => Promise<string>
}
```

Does it make sense to convert the whole API to Promise/async? Sync' transpiler can still return string directly without promise.

sverweij:
Today I've run a small test with the cli against a code base with some svelte in it, and it currently doesn't detect any dependencies in .svelte files. Root cause is indeed transpile now returning a Promise for svelte files. The calling code currently expects a string and doesn't handle promises:

```javascript
const transpile = require("./src/extract/transpile");

console.log(transpile(".svelte", "<script>const hello=42</script>"));
// returns Promise { <pending> }, whilst we'd expect a javascript program

```

### Solution direction: make the transpile step async?

Not now.

As to your suggestion to make the transpilation step async. That would also mean the _extract_ step, and with that _main_  which is not a small change. I do want to try to wet my feet with async some day, b.t.w., just to see if claims about better performance also hold for a single threaded cli app (sceptic, but willing to be convinced otherwise), but I'd rather manage that change (+ adapting the lion's share of the test suite) outside of this PR in a period where I have some more bandwidth available for big changes.


### Solution direction (wild thought): make the svelte compiler's preprocessing sync

I did have look at svelte's preprocessor,
- the svelte format is relatively straightforward
- the svelte's preprocessor is currently [regexing its way through the .svelte](https://github.com/sveltejs/svelte/blob/master/src/compiler/preprocess/index.ts#L165) and replacing everything between <script></script> tags with whatever the script you pass it returns.

Based on this I made a little PoC with a sync svelte preprocessor (that only does only what we need it for, so no sourcemap fancyness for instance). It seems to be holding up quite nicely with whatever I've thrown at it so far.

A solution like that can only serve temporary until either svelte grows a sync preprocessor or dependency-cruiser gets 100% async, though - think I need to sleep a night over it.

truongsinh:
I'd say having our own svelte preprocessor would be unsustainable in the long run, so I'd prefer dependency-cruiser gets 100% async. But I totally getting depcruise async is way out of scope for this PR.

sverweij:
> I'd say having our own svelte preprocessor would be unsustainable in the long run

Time will tell. I've put a few unit tests in place that compare outcome of the drop in sync preprocessor with svelte's async one. When svelte's async preprocessor changes behaviour in any way significant to our use, it'll show up in the automated tests.

sverweij (approved):
GEFGP

PR #424, opened 2021-01-17
perf: limit external manifest resolution to positively external modules only

sverweij:
## Description

When resolving external manifests (package.json's somewhere in the node_modules black hole) dependency-cruiser already prevented that for local or core modules. With this PR it exclusively does that for modules that it's positive are external dependencies. 

=> There's a slight difference in behavior for AMD modules between this PR and what went before. As there's no node-style resolution happening for AMD, it won't find unqualified external modules. So e.g. requiring `snodash` typically resolves to a local module. Before this PR, dependency-cruiser erroneously _did_ try to find `snodash`'s `package.json` somewhere in node_modules. After merging this PR this will no longer be the case.

## Motivation and Context

Resolves are fairly resource intensive, so each one we can save makes dc faster.

## How Has This Been Tested?

- [x] green ci
- [x] statistically significant performance tests (with`non run test:load`) showing an improvement on the benchmark on macOS 10.15.7 on a 2,6 GHz Quad-Core Intel Core i7 of ~50ms. Gain is likely more on slower machines (and on MS windows).

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


sverweij:
pulled too early

PR #438, opened 2021-02-06
fix bug where baseDir is ignored when calling programmatically

gaggle:
Running a scan with cruiseOptions `baseDir` set almost works, but actually results in various errors and bad scenarios because the run always sees `fileDirectoryArray` as relative to `process.cwd()` instead of the path specified in `baseDir`.

## Description

When calling dependency cruiser programmatically my understanding is specifying `baseDir` in cruiseOptions should be as if I ran `depcruise` in the shell from that folder. But `gatherScannableFilesFromDirectory` breaks that because it calls `readdirSync(pDirectoryName)` which *implicitly* relies on cwd when passed a relative path. If we instead join `pOptions.baseDir` into that path then `pDirectoryName` gets resolved according to `baseDir`.

By default `baseDir` is cwd so this doesn't change existing functionality, but allows the caller to override baseDir via options.

The same problem exists on L38 where stats are called.

With this fix in place I can now call this:
```js
futureCruise(
  ['.'],
  { outputType: 'dot', baseDir: dirOfMyChoosing },
)
```
Which generates the same output as if I run:
```bash
dirOfMyChoosing $ depcruise --output-type dot .
```

### Possible more work to do

I noticed Typescript definition for `ICruiseOptions` doesn't include `baseDir` so it may be I'm violating the intentions of the API. In that case I'd love to hear if there's anything I can do to salvage the intention of my PR which is to have `futureCruise` generate output relative to a specified folder.

## Motivation and Context

API caller can now specify `baseDir` and have the `fileDirectoryArray` be resolved relative to that.

## How Has This Been Tested?

I've only verified my own output now works, and have run the shell command and API call under debugger side-by-side to compare the function calls and variables resolve identically.

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/doc/options-reference.md
+++ b/doc/options-reference.md
@@ -653,6 +653,13 @@ favourite editor. Here's an example for visual studio code:
 
 > :bulb: Make sure the prefix ends on a `/`.
 
+### `baseDir`: specify a directory to cruise from
+
+> :shell: there is no command line equivalent for this at the moment
+
+By default dependency-cruiser will take the current working directory to start
+a cruise from. If you want to alter that you can pass it in this attribute.
+
 ## reporterOptions
 
 In the `reporterOptions` attribute you can pass things to reporters to influence
--- a/src/extract/gather-initial-sources.js
+++ b/src/extract/gather-initial-sources.js
@@ -27,7 +27,7 @@ function shouldNotBeExcluded(pFullPathToFile, pOptions) {
 
 function gatherScannableFilesFromDirectory(pDirectoryName, pOptions) {
   return fs
-    .readdirSync(pDirectoryName)
+    .readdirSync(path.join(pOptions.baseDir, pDirectoryName))
     .map((pFileName) => path.join(pDirectoryName, pFileName))
     .filter((pFullPathToFile) =>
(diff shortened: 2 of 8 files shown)


gaggle:
(BTW thanks for making this package public, it looks like a lot of work. The code is nice to look through)

sverweij:
Hi @gaggle thanks for this PR.

> I noticed Typescript definition for `ICruiseOptions` doesn't include `baseDir` so it may be I'm violating the intentions of the API.

It wasn't part of the API yet because it didn't work (yet) - couldn't find the root cause and once I realised the cli didn't really need it, I put fixing it on the back burner. So I'm happy you found the fix. When it works, the baseDir should become part of the ICruiseOptions and the JSON schema as well and be documented (added to the PR - see 676d4db and 03f9ae3)

> I've only verified my own output now works, and have run the shell command and API call under debugger side-by-side to compare the function calls and variables resolve identically.

To prevent  accidental regressions in the future this PR should have some unit tests. I've added two (in e0e9565) that should suffice for now. 

> (BTW thanks for making this package public, it looks like a lot of work. The code is nice to look through)

Thanks!

sverweij:
Released as just now as part of `dependency-cruiser@9.23.0`

PR #460, opened 2021-05-04
correctly filter out dependents from pathNot

davidparkagoda:
It looks like the minimum/maximum dependents count does not correctly filter from `from.pathNot`. Specifically I was getting weird results when I was trying to find files only referenced by test files.

> do that.

## Description
I negated the pathNot matching in the dependents filtering logic.

## How Has This Been Tested?
For some reason I cannot wrap my head around the tests that already exist so I made my own simple test that is not included in this pr.
```const USED_BY_APP = {
  from: {
    pathNot: ".test.ts$",
  },
  module: {
    path: "^src",
    numberOfDependentsLessThan: 1,
  },
};

  it("should complain about files that are only used by tests", () => {
    expect(
        matchesDependentsRule(USED_BY_APP, {
          source: "src/index.ts",
          dependents: [
            "src/index.test.ts",
          ],
        })
    ).to.equal(true);
  });

  it("should does not complain about files that used by app", () => {
    expect(
        matchesDependentsRule(USED_BY_APP, {
          source: "src/component.ts",
          dependents: [
            "src/index.ts",
          ],
        })
    ).to.equal(false);
  });

  it("should does not complain about files that used by both app and test", () => {
    expect(
        matchesDependentsRule(USED_BY_APP, {
          source: "src/component.ts",
          dependents: [
            "src/index.ts",
            "src/index.test.ts",
          ],
        })
    ).to.equal(false);
  });
```

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/src/validate/match-module-rule.js
+++ b/src/validate/match-module-rule.js
@@ -51,7 +51,7 @@ function dependentsCountsMatch(pRule, pDependents) {
   const lMatchingDependentsCount = pDependents.filter(
     (pDependent) =>
       Boolean(!pRule.from.path || pDependent.match(pRule.from.path)) &&
-      Boolean(!pRule.from.pathNot || pDependent.match(pRule.from.pathNot))
+      Boolean(!pRule.from.pathNot || !pDependent.match(pRule.from.pathNot))
   ).length;
   return (
     (!pRule.module.numberOfDependentsLessThan ||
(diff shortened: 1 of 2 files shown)


sverweij:
Thanks for this PR @davidparkagoda - good catch and the code is looking good, including the unit tests you did adapt. I’ll change the fixtures in those to something people who don’t speak Dutch will understand as well.

davidparkagoda:
Cool

I am actually okay with the dutch words in the tests. What was confusing me was that there is no intuitive reading related to snackbars and utensils.

Maybe the "dont share tests" can use people. I do not want people to share utensils 😄

PR #470, opened 2021-06-21
bugfix(cli): fix --exit-code flag

neelance:


sverweij:
thanks for the PR @neelance !

Seems I forgot a spot when upgrading commander some time ago. I'll put something on the backlog to ensure this becomes part of the integration test suite so it won't happen anymore.

(I'll merge & release tonight Dv)

neelance:
Fix deployed. Thanks for the quick release! 👍

Issue #490, opened 2021-08-30
How to exclude `type` imports from being processed by `forbidden` rule


agomai:
### Summary

Is there a way to exclude TypeScript `type` only imports from being considered in a `forbidden` rule?

### Context

I have a rule where I restrict a module from importing anything but some allowed modules, but I want that it is still allowed to import `type` definitions from these modules.

In my specific case I want a skeleton component only to be allowed to import from some allow-listed modules to prevent bundle size bloats. Importing types should be allowed as these don't affect bundle size

~~~javascript
{
    name: 'skeleton-only',
    severity: 'error',
    from: {
      path: ['.+-skeleton\\.tsx$'],
    },
    to: {
      dependencyTypes: ['local', 'localmodule'],
     // is this the way?
     // preCompilationOnly: true,

      // forbid all…
      path: [`^src/.*`],
      // …allow:
      pathNot: [
        // other skeletons
        `.+-skeleton.*\\.tsx$`,
       // others excluded for brevity
      ],
    },
  }
~~~

Currently I configured `options` `tsPreCompilationDeps: true` as I generally want Typescript dependencies to be considered.
I tried `tsPreCompilationDeps: specify` with `preCompilationOnly: true` in the rule and a type import from a forbidden module was not considered (the desired behavior). However this was also the case with `preCompilationOnly: false`, so I'm not sure if this works right. Setting `preCompilationOnly: undefined` caused no violations to be found at all (both type and regular imports) for the rule (bug?). Only not setting `preCompilationOnly` at all caused the `type` imports to be causing errors again  – not the desired behavior, but I tried as a negative test and to find out what `tsPreCompilationDeps: specify` does by default as I want all other rules to still consider type imports.

So questions are: 

1. How do I exclude TypeScript `type` only imports from being considered in a `forbidden` rule?
  * Is there are feature to explicitly target `type` definitions which I can enable/disable per rule or do I have to use `tsPreCompilationDeps` with `preCompilationOnly`
2. If `tsPreCompilationDeps` with `preCompilationOnly` is the way:
  * Which config do I have to use to only disable considering `type` imports for that rule but still keep other rules to consider TypeScript
  * What is the default behavior for `tsPreCompilationDeps: specify`? Does a rule consider TypeScript by default if `preCompilationOnly` is not set?
  * Why does `preCompilationOnly: true/false` seem behave the same way?
  * Why does `preCompilationOnly: undefined` caused no violations to be found at all, is this a bug?

[...]

PR #499, opened 2021-09-18
feature(extract): label typescript type-only imports as such

sverweij:
## Description

- Adds the `type-only` as a possible dependencyType so it can be used in rules

## Motivation and Context

Addresses a thing discovered while answering #490 - so ...

```typescript
import type { IHelloThingType } from "./things";
```

... is not only classified as a `localmodule` (or whatever else it is), but also as `type-only`, which then can be used as a filter in rules.

Caveats:
- Only works with TypeScript and only when dependency-cruiser's `tsPreCompilationDeps` option is switched to either `true` or `"specify"`...
- ...and only when using MicroSoft's TypeScript compiler (which is the default anyway when it's present and the source is TypeScript. Support for using this with `swc` might follow - provided `swc` doesn't discard these things right of the bat (not researched yet).

## How Has This Been Tested?

- [x] automated non-regression tests; green ci
- [x] additional unit tests
- [x] manual sanity tests

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


rklos:
There is something I don't understand or this feature doesn't work. I prepared super simple example with one file: https://github.com/rklos/dependency-cruiser-issue-example

index.ts has two type imports:
```ts
import type { Context as Ctx } from '@nuxt/types';
import type { Inject } from '@nuxt/types/app';
```

In dependency-cruiser's config I have one single rule
```js
{
  from: {},
  to: {
    dependencyTypes: [
      'type-only',
    ],
  },
},
```

And after calling
```bash
depcruise --config .dependency-cruiser.js .
```
I get "no dependency violations found" instead of expected warnings about forbidden type imports.

sverweij:
@rklos thanks for checking this, and for the repro repo. You totally hit a bug. Type only detection works, but it isn't passed on correctly to the spot that matters. Never trust 100% test coverage :-) 

(PR will be updated with a fix)

sverweij:
fixed version (the tests of which now also has an integration test that checks for non-regressions) published as `dependency-cruiser@10.4.0-beta-2`

PR #519, opened 2021-10-18
Adding Vue3 @vue/compiler-sfc support

annamooseity:
If you are using Vue3, you might include packages that are incompatible with vue-template-compiler, causing the "isAvailable" function for Vue SFCs to return false. Now, if vue-template-compiler fails to load, we go ahead and try @vue/compiler-sfc as well, in case we have that instead..

## Description

Added a different Vue single file component transpiler, following the pattern of what was done with coffeescript modules. 

## Motivation and Context

I want to use dependency-cruiser with my project that includes Vue3 files, but when it checks for available transpilers, `require`ing "vue-template-compiler" errors out because of incompatibility. I added a compatible Vue 3 single file component transpiler that is used if the first one doesn't work.

## How Has This Been Tested?

All existing Vue unit tests still pass. I've also added a new test file, vue3-template-wrap.spec.js, that runs the same tests as the existing file, but it "hides" the "vue-template-compiler" which forces the code to use the Vue3 transpiler. Thanks @sverweij for your help with this!!

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -298,9 +298,11 @@ situations without workarounds).
 
 **A**: Yes.
 
-For `.vue` single file components it uses the `vue-template-compiler` (which will
-be in your module dependencies if you're developing with Vue - just make sure
-you install dependency-cruiser in the same spot).
+For `.vue` single file components it uses _either_ the `vue-template-compiler` (Vue2) or `@vue/compiler-sfc` (Vue3).
+
+`vue-template-compiler` should be in your module dependencies already if you are developing with Vue2 - just make sure you install dependency-cruiser in the same place.
+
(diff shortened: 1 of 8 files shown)


sverweij:
> @sverweij sounds good! I noticed that when I run my test individually, the proxyquire works (and therefore coverage is okay) - but when I run `yarn:test` the proxyquire trick doesn't seem to work (so no coverage of Vue3 stuff). I'll be investigating this today - I probably need to set up some additional mocha settings or something. Let me know if you know of anything off the top of your head! Thanks!

On the ci the new tests _do_ get run [e.g here] (https://github.com/sverweij/dependency-cruiser/runs/3966756342?check_suite_focus=true#step:9:1071); no clue why it wouldn’t ‘unload’ the older vue module.

sverweij:
Back at a keyboard again :-) I've given the code a spin locally and it works 👌 . 

Thanks again for this (very complete) PR @annamooseity, and also for your patience!

annamooseity:
Yay!!!! Thanks for all your support!!

PR #547, opened 2021-12-25
fix(report): adds collapsePattern to metrics report

vedrani:
## Description

Metrics report doesn't support collapse pattern, so we cannot collapse results. I used existing report API that contains `collapsePattern` to get regex for filtering components of report.

## Motivation and Context

I tried to use [depcruise-fmt](https://github.com/sverweij/dependency-cruiser/blob/develop/doc/cli.md#depcruise-fmt) to get various results. Once I tried `metrics` report with options like `focus`, `include-only` it didn't work as I expected. I would just get original file in different format (json > table). For example:

`yarn depcruise-fmt -T metrics -I "^(hosts)/[^/]+$"  metrics.json -f metrics.txt`

I expected I will be able to filter or include only rows that represent accumulated metrics for top-level folder.

## How Has This Been Tested?

I tested manually on my example, but I can add tests.

`yarn depcruise-fmt -T metrics --collapse "^(hosts)/[^/]+$" metrics.json -f metrics-shorted.txt`

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


src/report/metrics.js
```diff
@@ -80,6 +80,11 @@ function transformMetricsToTable(
   lComponents = lComponents.concat(
     hideModules ? [] : modules.filter(metricsAreCalculable).map(metricifyModule)
   );
+    
+  const collapsePatternRegex = new RegExp(collapsePattern, 'g');
+  lComponents = lComponents
+    .filter(({ name }) => name.match(collapsePatternRegex))
```
sverweij, line 86: Although it takes a collapsePattern it performs a filter and not a collapse/ aggregation (which would have to consolidate module counts, Ca/ Ce and re-calculate instability - which currently not even the module [consolidation](https://github.com/sverweij/dependency-cruiser/blob/develop/src/graph-utl/consolidate-to-pattern.js) takes into account). 

It seems the intent of this code is more like what --include-only (/ includeOnly) is for. For command line use with `depcruise-fmt`/ the reporter API (as indicated in the PR description) adding a folder filter in [main/report-wrap.js](https://github.com/sverweij/dependency-cruiser/blob/develop/src/main/report-wrap.js) then might be a better fit. As a bonus other reporters that take folders into account will then benefit as well.

sverweij:
Hi @vedrani thanks for this PR! 

The filter options of depcruise-fmt (_include-only_, _exclude_, _focus_) and the _collapse_ option works on all reporters in the same fashion (centralised in [main/report-wrap.js](https://github.com/sverweij/dependency-cruiser/blob/develop/src/main/report-wrap.js#L11)). It currently only on works on modules (although collapse doesn't consolidate metrics correctly yet), but not on folders - a fact I hadn't realised before you opened this pull request, so I'm happy you raised it.

There's a few notes regarding the code in this PR:
- This improvement should probably primarily be placed in [main/report-wrap.js](https://github.com/sverweij/dependency-cruiser/blob/develop/src/main/report-wrap.js) - that way all current and future reporters will benefit from it, instead of just the metrics one. 
- From the description in the PR and the actual change it looks like the intent is to have a filter - like _include only_ and not a collapse/ aggregation one. Did I infer that correctly?
- On adding tests: yes please :-). For bug fixes & new features it helps repeatably prove the code is doing what it intents/ explains what it's for and it'll prevent future regressions.
- The ci flagged a code formatting issue - doing an `npm run lint:fix` will likely fix that.

vedrani:
> * This improvement should probably primarily be placed in [main/report-wrap.js](https://github.com/sverweij/dependency-cruiser/blob/develop/src/main/report-wrap.js) - that way all current and future reporters will benefit from it, instead of just the metrics one.

Thanks for advice, I added filtering for folders. Can you give me advice is this good direction? I need additionally to check `focus` for folders.

> * From the description in the PR and the actual change it looks like the intent is to have a filter - like _include only_ and not a collapse/ aggregation one. Did I infer that correctly?

Yes, seems metrics reporter already calculates correct values for top-level folders and I only want to output top-level.

> * On adding tests: yes please :-). For bug fixes & new features it helps repeatably prove the code is doing what it intents/ explains what it's for and it'll prevent future regressions.

Will add them next.

PR #556, opened 2022-01-10
feature(extract): add vue script setup support

bashess:
## Description

Vue Script Setup was not supported. So i add a check for script setup on the extraction.

## Motivation and Context

Cant use Dependency Cruiser with new Vue setup syntax

## How Has This Been Tested?

Tested with private project and add tests

## Types of changes

- [X] New feature (non-breaking change which adds functionality)

## Checklist

- [X] :book:

- [X] :balance_scale:


sverweij:
hi @bashess, thanks for this pull request! Really appreciate PR's like this, as it's hard for me to keep up with things happening in all frameworks dependency-cruiser supports. There's also a good additional automated tests going on, which is super helpful.

Two questions:
- Does vue allow SFC's with both a `scriptSetup` and a `script` element? Because in that case we might want to handle that situation.
- This PR updates the function for vue3, but leaves the one for vue2 alone (in [vue-template-wrap.js#L46](https://github.com/sverweij/dependency-cruiser/blob/develop/src/extract/transpile/vue-template-wrap.js#L46). Do `scriptSetup` elements only occur in vue3?

And a very small thing: the ci  complains about some linting/ formatting. This can likely be fixed by running `npm run lint:fix` locally. I'd have expected husky/ lint-staged to catch that, but apparently that didn't fire on your local machine?
> The script setup in package.json is a bit unconventional in that the `test` target only runs the test and that
> there's a `check` and a `check:full` target that in ~40s runs all checks (tests, linting, formatting etc).

src/extract/transpile/vue-template-wrap.js
```diff
@@ -40,6 +40,12 @@ function vue3Transpile(pSource) {
     return "";
   }
 
+  const isSetup = Boolean(_get(parsedComponent, "descriptor.scriptSetup", ""));
+
+  if (isSetup) {
+    return _get(parsedComponent, "descriptor.scriptSetup.content", "");
+  }
+
```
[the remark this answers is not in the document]
bashess: You are totally right, you can use both. I have pushed some changes and tests about that.

bashess:
> 

I would only support script setup for vue 3, vue 2 its only possible with plugins https://github.com/antfu/unplugin-vue2-script-setup

src/extract/transpile/vue-template-wrap.js
```diff
@@ -11,48 +11,39 @@ const { supportedTranspilers } = require("../../../src/meta.js");
...
+	}
 
-  if (!isEmpty(errors)) {
-    return "";
-  }
+	const scriptContent = _get(parsedComponent, "descriptor.script.content", "");
+	const scriptSetupContent = _get(parsedComponent, "descriptor.scriptSetup.content", "");
 
-  return _get(parsedComponent, "descriptor.script.content", "");
+	return scriptContent + scriptSetupContent;
```
sverweij, line 39: Does the order in which the concatenation happens matter?

sverweij, line 39: Let's say scriptContent is `const lThing = 123` and scriptSetupContent `import {hello} from 'some-module'` (both syntactically valid) - the concatenated code would look like this (which isn't valid):

```javascript
const lThing = 123import {hello} from 'some-module'
```

... unless the template compiler ensures there's a line break at the end of each of them. 

=> Would inserting a line break (os.EOL) mitigate this problem sufficiently?

sverweij:
Hi @bashess thanks for the updates! I've added some inline comments

Apart from those it seems your editor reformatted the source code a bit - which is what makes the ci fail in the [linting step](https://github.com/sverweij/dependency-cruiser/runs/4802236012?check_suite_focus=true#step:7:27). Could you run an `npm run lint:fix` before pushing next changes? It'll also make it easier to compare the new and old code. Normally husky + lint-staged should take care of that in a pre-commit hook, but that doesn't seem to happen on your machine.

src/extract/transpile/vue-template-wrap.js
bashess: done

bashess: I think that shouldn't matter

bashess:
> 

Doing lint:fix i get ~170 file changes and i have to pass "no-useless-concat": "off" in eslintrc to get it work without errors. Is this a local issue on my side? Im sorry about that i dont get it to work to pass all checks

bashess:
> > 
> 
> Doing lint:fix i get ~170 file changes and i have to pass "no-useless-concat": "off" in eslintrc to get it work without errors. Is this a local issue on my side? Im sorry about that i dont get it to work to pass all checks

So i hope all looks normal now. Sry about that, and the force push. I had in a higher folder a prettierrc, that was the problem.

sverweij:
LGTM - thanks again for the PR. Will try to make a release in the coming days that includes this feature.

sverweij:
released a few minutes ago as part of release v11.3.0

PR #566, opened 2022-02-06
feature(tool): support "zoom" interactions in tool `depcruise-wrap-stream-in-html`

zthxxx:
## Description

Add support for zooming interactions in tool `depcruise-wrap-stream-in-html`

Zoom is a natural way to interact in most popular app:
- using mouse wheel scroll with hold <kbd>Ctrl</kbd>/<kbd>Command</kbd> key
- or two-fingers pinch in trackpad

Also can move viewport by two-fingers pan in trackpad, or mouse wheel scroll (with <kbd>Shift</kbd> key for reverse)

## Motivation and Context

Most time we use `dependency-cruiser` to analyze some **complex and huge repo**, and also need the interaction ability like "highlight by hover" or "click and jump link" to help our analysis, so we must choose the **`depcruise-wrap-stream-in-html`** tool.

However, the viewport of content is very difficult to use, due to content is too large and viewport is too small, the **scroll** is the only thing we can do.

That is lack **zoom** interactions,

so we can't see some part of level or the whole content, only see details of a little files or lines, like that below:

image

(but the whole svg like this below:)

image

## How Has This Been Tested?

The demo file generated by this command below, you can checkout the branch and try, 

or just see the file which I uploaded in https://dependency-graph-zoom-demo.vercel.app

```bash
npx depcruise -v -T dot src | dot -T svg | npx depcruise-wrap-stream-in-html > dependency-graph-zoom-demo.html
```

## Screenshots

https://dependency-graph-zoom-demo.vercel.app

image

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


sverweij:
Hi @zthxxx - thanks for this pull request. It seems to be a complete re-write of the javascript and I thank you for the effort that went into it. I won't be able to merge the PR as is, though, even thought there's good functional bits in there - see below for my reasoning.

- apart from what the PR describes there seems to be some additional functionality in it
  - the highlight line widths have changed 
     it used to be 1.5x line width - to make it distinct from regular edge and node line widths. This PR reduces it to 1.2x line width (stroke-width 2.4). Combined width the more subdued colors the distinction from normal edges and nodes did become less.
  - the color of incoming/ outgoing edges changed
  from fuchsia for everything to green for incoming and blue for outgoing. The distinction between incoming and outgoing is a good addition. The two colors the PR picks do not seem stand out much, though, especially when the regular colouring already uses these colors (blue and green will be common choices in color schemes). Maybe this will be less of a problem when 
  - clicks on nodes now  override browser behaviour for links
  instead of navigating to the file in the same tab, it opens a new tab/ window. This seems like a good idea at first, but  if they want to open in new tab they'll use the browser's functionality for that (e.g. ctrl-click). I'm also not sure how this would fare with custom URL schemes (e.g. `vscode://` which opens a link in vscode).
- on the zoom functionality in this PR:
  - a nice thing: it makes the zoom continuous and enables to go beyond the minimum and maximum zoom levels of the browser.
  - a drawback of the current approach is that it blocks scrolling/ panning on touch-only devices (tested on an android phone and an iPad)
  - another drawback one is that keyboard navigation doesn't seem to work anymore (tried page up & down, home, end, arrow keys).
  - I also noticed that when scrolling it's easy to end up in a white area far outside the graph (even on top)
  - given this, the additions this PR makes to the browsers built-in zoom (ctrl-+, ctrl--, ctrl-0) currently don't outweigh the drawbacks of the implementation.
- from a code perspective:
  - it changes the 'concat three files together' approach to a templated one. Theoretically this'd limit the maximum size of the svg and might be slightly slower - but on the  that against better readability (+ the separate files have more clear 'responsibilities'); I think it's a good choice.
  - the PR changes original approach of the javascript part of the code - but to me it's not clear why it does so.
  - the original javascript used 'old' guaranteed-to-be-working-in-all-browsers constructs only (e.g. `var` over `const` & `let`, no classes). Maybe the browser landscape has moved on since and it might actually be a  legit choice; just wasn't expecting this (theoretically breaking) change in a PR like this.
  - as a minor gripe: the code style doesn't adhere to the one in use in the repo. This is partly my fault as  the style checker (and linter) didn't fire properly on the CI or the pre-commit trigger (partly fixed in b83cca827e21cce3a8adfc5afbf3f3cf74e9b7b7).

PR #577, opened 2022-04-02
fix(readme): anchor in readme

zi-gae:
## Description

Just some minor changes to replace the href in anchor.

## Motivation and Context

replace to href


sverweij:
thanks for this patch @zi-gae !

Issue #353, opened 2020-08-20
Feature request: Add a mermaid reporter


dagadbm:
### Context

It would be nice if the dependency graph could be generated using mermaidJS (assuming its possible)

### Expected Behavior

Output a .md file with the mermaid syntax on it
```mermaid
graph
 ...
 ...
```

dagadbm:
I guess the main "problem" would be to drop the graphviz dependency and have the graph part of the documentation as text (gitlab even reads mermaid for you automatically as well)

Im guessing the flowchart one (https://mermaid-js.github.io/mermaid/#/flowchart) would probably the most similar one to what this tool does.

i am testing this tool out for now but I was surprised for it to not support such a well known (at least it seems like it) project.

sverweij:
For anyone who has an interest in picking this up - dependency-cruiser now has a plugin mechanism for reporters that should make developing and maintaining reporters easier - see the [_How do I add a new output format_](https://github.com/sverweij/dependency-cruiser/blob/develop/doc/faq.md#q-how-do-i-add-a-new-output-format) section in the FAQ for a starting point.

fabiospampinato:
> To research: does it scale to ~200 nodes & ~300 edges?

Maybe, but it would take a while to generate the diagram for thousands of nodes, mermaid is very slow in my experience, it touches the DOM too much.

sverweij:
I guess that with #599 this issue can be considered as _fixed_. For future visitors; it's currently implemented as a plugin, which can be invoked e.g. thusly:

```sh
depcruise src --config .dependency-cruiser.js --output-type plugin:dependency-cruiser/mermaid-reporter-plugin
```

It isn't as feature rich as the the `dot` reporters - and _can't_ be due to limitations in mermaid (see [comment above](https://github.com/sverweij/dependency-cruiser/issues/353#issuecomment-678693190)). But if the goal is to visualise the dependency-graph, without highlighting detected errors, warnings, cycles etc. this might serve.

PR #608, opened 2022-06-05
ci(gh-actions): use dependency-cruiser-report-action on pull request

MH4GF:
## Description

Add workflow of https://github.com/MH4GF/dependency-cruiser-report-action/ .
Visualize dependencies of changed files and add comments. Made with plugins added by This PR https://github.com/sverweij/dependency-cruiser/pull/599 .

## Motivation and Context

First off, thanks for making a cool tool!
I have created an action to add a diagram visualized using dependency-cruiser as a PR comment. 
I think this will be useful for discussions on finding inappropriate dependencies and considering better architectures.

## How Has This Been Tested?

Here is a demo: https://github.com/MH4GF/test-depcruiser/pull/1#issuecomment-1146736834

This workflow has not yet been activated. This is because `pull_request_target` is used to trigger the workflow, which will not fire unless the yaml file exists in the default branch. ref: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


.github/workflows/visualize-deps-of-changed-files.yml
```diff
@@ -0,0 +1,17 @@
+name: visualize dependenices of changed files in pull request
+on:
+  pull_request_target:
```
MH4GF, line 3: The reason for using `pull_request_target` instead of pull_request, the PR at the forked cannot reference the GITHUB_TOKEN of the fork source.
`pull_request_target` should generally be treated with caution from a security standpoint, but the reason `pull_request_target` was originally created was to add a comment to the PR. 

> The reason to introduce the pull_request_target trigger was to enable workflows to label PRs (e.g. needs review) or to comment on the PR. The intent is to use the trigger for PRs that do not require dangerous processing, say building or running the content of the PR.

ref: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

Since this purpose also does not involve hazardous processing by simply adding a comment, it would seem to make sense to use `pull_request_target` .

MH4GF:
@sverweij I'm so glad you put so much effort into this! Thank you.

Using the focus option seems like a very good idea! As you pointed out, it's better to show the files around the change diff for better understanding, and there was a problem with not knowing where the root file was in the bigger picture, such as deep dependencies. I'll try to test and implement this in action afterwards to replace the focus option. 

> There will be pull requests that don't touch source code - but instead update documentation. How does the action handle that?

```
depcruiser -I src -T plugin:dependency-cruiser/mermaid-reporter-plugin README.md
```

In this case, the exit code is 0 and the action just displays it as it is, so it will look like this: https://github.com/MH4GF/dependency-cruiser-report-action/pull/53#issuecomment-1146741803

I noticed that you pointed this out, I'd like to show that there were no files to display, as it's hard for the user to see what's going on. 
If there is a cli option in dependency-cruiser to return exit 1 if there are no files to show, it might be useful when running with CI like this action or https://github.com/sverweij/dependency-cruiser/pull/614 (this feature very nice too!). 

> In https://github.com/MH4GF/dependency-cruiser-report-action/issues/52 it seems the action bumps into a situation where it feeds dependency-cruiser some files it cannot handle

Oh, thanks for looking into all that! I'll fix this on the action side so that only supported extensions are passed to dependency-cruiser.

> I'm not sure of mermaid's performance with large change sets (e.g. see https://github.com/sverweij/dependency-cruiser/issues/353#issuecomment-830050992 in issue https://github.com/sverweij/dependency-cruiser/issues/353)). If this indeed is an issue, would it be useful to put an upper limit to the number of files to report on?

It may be constant useful, but I rather think the problem is not the slow rendering in case of large dependencies (although the slowness is certainly felt), but the inability to render. mermaid.js has a default text character limit of 50000 characters, and if you exceed it, an error message will be displayed. If you run the following script in the dependency-cruiser repository, it will be about 110000 characters.

```
node bin/dependency-cruise.js  -T plugin:dependency-cruiser/mermaid-reporter-plugin src bin test configs types tools --config
```

In this case, the user would have to adjust doNotFollow/exclude/maxDepth, etc. I will also take time to think about what could be done on the dependency-cruiser side.

sverweij:
whoops @ stalebot - wasn't done with this PR yet.

PR #613, opened 2022-06-06
fix(mermaid-plugin): replace some meta chars

### LINKED ISSUE #612: Issue(mermaid-reporter-plugin): Mermaid.js fails to render when some symbols are included in the file name
### Expected Behavior

Can render.

### Current Behavior

Cannot render if the following symbols are included.

```
// If followed by two hyphens.(single is not a problem)
foo--bar[foo--bar]

// If the alias could not be resolved
~/baz["~/baz"]
```

### Possible Solution

I think both can be replaced to underscore.

- While a single hyphen is fine, if more than one follows, it is an error, so we consider all hyphens to be acceptable replacements. Readability is not easily compromised.
- If the alias cannot be resolved, it will appear as a separate directory, but consider it better than not even rendering.

MH4GF:
## Description

Closes #612

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


sverweij:
Hi @MH4GF thanks again! 

These characters might not be the last we encounter - but they'll for sure cover many cases already.

PR #616, opened 2022-06-07
feature(mermaid-plugin): adds focus highlighting

### LINKED ISSUE #564: Feature request: Focus with depth support
### Context

When currently using `--focus` option. It only shows the direct dependencies and dependents. In practice, I have been using `focus` to figure culprits which break code splitting (and hence lazy loading).

Since, I want to follow the chain of dependents to a large size module which I want to lazy load a depth option will be helpful.

### Expected Behavior

`--focus-depth N`  when using `--focus`.

This makes the output contain upto `N` dependents/depedencies from the focused modules.

### Current Behavior

`--focus` only shows direct dependents/dependencies.

### Considered alternatives

Currently I have to recursively call focus manually on the chain, to figure out the root culprits which unnecessary effort.

sverweij:
## Description

- adds focus highlighting to the mermaid reporter;

Also includes the mermaid reporter plugin along with the regular test coverage (it was already 100% covered to start with - and we don't want that to grow less ;-) )

## Motivation and Context

@MH4GF has been working on a _very nice_ [GitHub action](https://github.com/MH4GF/dependency-cruiser-report-action) that can add a mermaid report to
your pull request that shows the modules the PR is about. It'd be awesome if they could expand a bit to not only show the modules touched by the PR, but _also_ the immediate context. This is what dependency-cruiser's (existing) focus feature is for. 

In the context of a PR it'd however be useful to see which modules are the ones _touched_ by the PR and which are included in the model for _context_ only. This PR adjusts that.

> This feature might profit from an implementation for #564; it might be useful
> to see even _more_ context than the direct dependents and dependencies of the modules
> at hand. 

## How Has This Been Tested?

- [x] green ci
- [x] additional unit test

## Screenshots

Given a dependency-cruiser result which has a `focus` set to one or more modules the output will show these modules highlighted (in green) and the modules in its context in a slightly subdued gray:

### After

Here a reviewer would see _normalize.js_ has been updated - but interestingly enough not any of the test files that directly include it.

```mermaid
flowchart LR

subgraph src["src"]
  subgraph src_main["main"]
    subgraph src_main_rule_set["rule-set"]
      src_main_rule_set_normalize_js["normalize.js"]
    end
    src_main_index_js["index.js"]
    subgraph src_main_utl["utl"]
      src_main_utl_normalize_re_properties_js["normalize-re-properties.js"]
    end
  end
end
subgraph test["test"]
  subgraph test_enrich["enrich"]
    subgraph test_enrich_derive["derive"]
      subgraph test_enrich_derive_reachable["reachable"]
        test_enrich_derive_reachable_index_spec_mjs["index.spec.mjs"]
      end
    end
  end
  subgraph test_main["main"]
    subgraph test_main_rule_set["rule-set"]
      test_main_rule_set_normalize_spec_mjs["normalize.spec.mjs"]
    end
  end
  subgraph test_validate["validate"]
    test_validate_parse_ruleset_utl_mjs["parse-ruleset.utl.mjs"]
  end
end
subgraph node_modules["node_modules"]
  subgraph node_modules_lodash["lodash"]
    node_modules_lodash_has_js["has.js"]
    node_modules_lodash_cloneDeep_js["cloneDeep.js"]
  end
end
src_main_rule_set_normalize_js --> src_main_utl_normalize_re_properties_js
src_main_rule_set_normalize_js --> node_modules_lodash_cloneDeep_js
src_main_rule_set_normalize_js --> node_modules_lodash_has_js
src_main_index_js --> src_main_rule_set_normalize_js
test_enrich_derive_reachable_index_spec_mjs --> src_main_rule_set_normalize_js
test_main_rule_set_normalize_spec_mjs --> src_main_rule_set_normalize_js
test_validate_parse_ruleset_utl_mjs --> src_main_rule_set_normalize_js

style src_main_rule_set_normalize_js fill:lime,color:black
```
> update 2022-06-08: replaced with diagram that _only_ highlights, and doesn't subdue

### Before

```mermaid
flowchart LR

subgraph src["src"]
  subgraph src_main["main"]
    subgraph src_main_rule_set["rule-set"]
      src_main_rule_set_normalize_js["normalize.js"]
    end
    src_main_index_js["index.js"]
    subgraph src_main_utl["utl"]
      src_main_utl_normalize_re_properties_js["normalize-re-properties.js"]
    end
  end
end
subgraph test["test"]
  subgraph test_enrich["enrich"]
    subgraph test_enrich_derive["derive"]
      subgraph test_enrich_derive_reachable["reachable"]
        test_enrich_derive_reachable_index_spec_mjs["index.spec.mjs"]
      end
    end
  end
  subgraph test_main["main"]
    subgraph test_main_rule_set["rule-set"]
      test_main_rule_set_normalize_spec_mjs["normalize.spec.mjs"]
    end
  end
  subgraph test_validate["validate"]
    test_validate_parse_ruleset_utl_mjs["parse-ruleset.utl.mjs"]
  end
end
subgraph node_modules["node_modules"]
  subgraph node_modules_lodash["lodash"]
    node_modules_lodash_has_js["has.js"]
    node_modules_lodash_cloneDeep_js["cloneDeep.js"]
  end
end
src_main_rule_set_normalize_js --> src_main_utl_normalize_re_properties_js
src_main_rule_set_normalize_js --> node_modules_lodash_cloneDeep_js
src_main_rule_set_normalize_js --> node_modules_lodash_has_js
src_main_index_js --> src_main_rule_set_normalize_js
test_enrich_derive_reachable_index_spec_mjs --> src_main_rule_set_normalize_js
test_main_rule_set_normalize_spec_mjs --> src_main_rule_set_normalize_js
test_validate_parse_ruleset_utl_mjs --> src_main_rule_set_normalize_js

```

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/configs/plugins/mermaid-reporter-plugin.js
+++ b/configs/plugins/mermaid-reporter-plugin.js
@@ -1,19 +1,18 @@
+/* eslint-disable security/detect-object-injection */
 const ACORN_DUMMY_VALUE = "✖";
 
-/* eslint-disable security/detect-object-injection */
-const mermaidNode = (pNode, pText) => {
-  const lNode = pNode
+const hashNodeName = (pNode) =>
+  pNode
     .replace(ACORN_DUMMY_VALUE, "__unknown__")
     .replace(/^\.$|^\.\//g, "__currentPath__")
     .replace(/^\.{2}$|^\.{2}\//g, "__prevPath__")
     .replace(/[[\]/.@~-]/g, "_");
-  const lText = pText ? `["${pText}"]` : "";
-  return `${lNode}${lText}`;
-};
+
+const mermaidNode = (pNode, pText) => `${hashNodeName(pNode)}["${pText}"]`;
 
 const mermaidEdge = (pFrom, pTo) => {
-  const lFromNode = mermaidNode(pFrom.node);
-  const lToNode = mermaidNode(pTo.node);
+  const lFromNode = hashNodeName(pFrom.node);
+  const lToNode = hashNodeName(pTo.node);
   return `${lFromNode} --> ${lToNode}`;
 };
 
@@ -84,6 +83,18 @@ const convertedSubgraphSources = (pCruiseResult) => {
   return lTree;
 };
 
+const focusHighlights = (pModules) => {
+  const lHighLightStyle = "fill:lime,color:black";
+
+  return pModules
+    .filter((pModule) => pModule.matchesFocus === true)
+    .reduce(
+      (pAll, pModule) =>
+        (pAll += `\nstyle ${hashNodeName(pModule.source)} ${lHighLightStyle}`),
+      ""
+    );
+};
+
 const renderMermaidThing = (pCruiseResult) => {
   const subgraphs = convertedSubgraphSources(pCruiseResult);
   const edges = convertedEdgeSources(pCruiseResult);
@@ -92,7 +103,7 @@ const renderMermaidThing = (pCruiseResult) => {
 
 ${mermaidSubgraphs(subgraphs)}
 ${mermaidEdges(edges)}
-`;
+${focusHighlights(pCruiseResult.modules)}`;
 };
 
 /**
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "test:i": "mocha --timeout 4000 \"test/**/*.spec.{js,mjs,cjs}\" -g \"^\\[[I]\\]\"",
     "test:u": "mocha --timeout 4000 \"test/**/*.spec.{js,mjs,cjs}\" --grep \"^\\[[U]\\]\"",
     "test:e": "mocha --timeout 4000 \"test/**/*.spec.{js,mjs,cjs}\" --grep \"^\\[[E]\\]\"",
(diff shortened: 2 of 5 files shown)


configs/plugins/mermaid-reporter-plugin.js
```diff
@@ -84,6 +83,22 @@ const convertedSubgraphSources = (pCruiseResult) => {
...
+    .reduce(
+      (pAll, pModule) =>
+        (pAll += `\nstyle ${hashNodeName(pModule.source)} ${
+          pModule.matchesFocus
+            ? "fill:lime,color:black"
+            : "fill:lightgrey,color:black"
+        }`),
+      ""
+    );
+};
```
MH4GF, line 100: Understanding that the color scheme is the same as that of the dot reporter, I think it would be better to add highlighting when `matchedFocus` is true and leave it as is otherwise. There are two reasons.

- mermaid.js and GitHub support dark mode. It seems better to have the background color change according to the user's environment instead of using lightgray as a uniform background color.
- The default value of mermaid.js and tools like GitHub and Notion seem to set the character limit at 50000, and it seems that you should reduce unnecessary characters as much as possible to avoid exceeding the maximum character limit.

sverweij: Agree, good point! I'll update it.

PR #626, opened 2022-06-26
bugfix(mermaid): small fixes to types, tests

sverweij:
## Description

Some things I forget to add (/ didn't notice in the PR review):

* bugfix(types): make mermaid optional in options type
* bugfix(test|mermaid): adds scenario where no options are passed to the reporter
* refactor(test|mermaid): dedupe comparison function
* refactor(mermaid): use .has instead of .get to test whether a key is in the hash map
* style(mermaid): apply some code style rules that don't have an automated check on them yet
  * these are things that don't matter for the functioning of the code, but do
    make it look a bit more familiar to the code style of the rest of the code base.
    As long as there's no automated check (or even better: automatic fix) on them
    I'm not comfortable to put these as review comments.
  * Replaces arrow function expressions with a regular named `function` where the function
    expression is a bit longer
  * Replaces shorter, single line function expressions with only `=> { return <somestatement>` } in theme with `=> <somestatement>`
  * Puts verbs in function names where this wasn't already done
  * Prefixes local variables/ constants with an `l` (eslint plugin budapestian missed these,
    which is an improvement that should go into that plugin).
  * Moves the hashNodeName closer to the function it's called from.

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/src/report/mermaid.js
+++ b/src/report/mermaid.js
@@ -75,19 +75,19 @@ const mermaidSubgraphs = (pSource, pOptions, pDepth = 0) => {
 
 /**
  * @param {import('../../types/dependency-cruiser').ICruiseResult} pCruiseResult
- * @param {Map<string, string>} pMinifiedNames
+ * @param {Map<string, string>} pNamesHashMap
  */
-const convertedSubgraphSources = (pCruiseResult, pMinifiedNames) => {
+const convertedSubgraphSources = (pCruiseResult, pNamesHashMap) => {
   let lTree = {};
 
   pCruiseResult.modules.forEach((pModule) => {
-    const paths = pModule.source.split("/");
+    const lPaths = pModule.source.split("/");
 
-    paths.reduce((pChildren, pCurrentPath, pIndex) => {
+    lPaths.reduce((pChildren, pCurrentPath, pIndex) => {
       if (!pChildren[pCurrentPath]) {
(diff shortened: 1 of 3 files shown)


src/report/mermaid.js
```diff
@@ -129,7 +129,7 @@ const hashModuleNames = (pModules, pMinify) => {
 
     for (let lIndex = 0; lIndex < lPaths.length; lIndex += 1) {
       const lName = lPaths.slice(0, lIndex + 1).join("/");
-      if (!lNamesHashMap.get(lName)) {
+      if (!lNamesHashMap.has(lName)) {
```
MH4GF, line 132: nice job! :)

test/report/mermaid/mermaid.spec.mjs
```diff
@@ -10,45 +11,60 @@ const requireJSON = createRequireJSON(import.meta.url);
...
+    same(
+      "dependency-cruiser-2019-01-14",
+      { minify: false },
+      mermaidReporterPlugin
+    );
+    same(
+      "dependency-cruiser-2019-01-14",
+      { minify: true },
+      mermaidReporterPlugin
+    );
```
MH4GF, line 68: 👍🏻

Issue #715, opened 2022-12-30
Feature request: cache based on content of files


7rulnik:
### Context

We already have [cache](https://github.com/sverweij/dependency-cruiser/blob/develop/doc/cli.md#--cache-use-a-cache-to-speed-up-cruising-experimental). It speeds up cruising massively. But it uses git to get modified files between runs. Usually, we have whole git tree, but sometimes we don't. For example, in CI people usually use git depth 1 for faster checkout ([gitlab for example](https://docs.gitlab.com/ee/ci/large_repositories/index.html#shallow-cloning)). So in this case we don't have git tree and can't use cache.

### Expected Behavior

It would be nice if we can use cache even if we don't have a whole git tree.

### Current Behavior

Currently, cache works only if we have SHA1 in git tree and if we don't we will run whole cruising from the start.

### Possible Solution

We can provide different strategies for caching or switch only for caching based on file content.
TypeScript uses file content for the [incremental feature](https://www.typescriptlang.org/tsconfig/incremental.html). Prettier, eslint, stylelint supports [--cache-strategy](https://eslint.org/docs/latest/user-guide/command-line-interface#--cache-strategy) flag. It allows choosing between strategies: modified time or hash of file content.

So on the first run, we can save the hash of file content into the cache.
And on the second we can go through modules, read them and compare actual hash of content with cached one.

So an implementation could be kinda like this:
```js
async function getChangedModules() {
    const changedModules = []
    for (const file of manifest.modules) {
        if (!file.couldNotResolve && !file.coreModule) {
            // We can use something like https://yarnpkg.com/package/p-limit here for concurrency
            const content = await fs.readFile(__dirname + '/' + file.source)
            const hashSum = crypto.createHash('sha256');
            hashSum.update(content);
            const hex = hashSum.digest('hex');
            if (hex !== file.hash) {
                changedModules.push(file.source)
            }
        }
    }

    return changedModules
}
```

On m1 pro 13 it takes 1s for 7200 modules.

So I think we can drop git for detecting changes and just use metadata and content like other tools or even always use content hashing stategy.

sverweij:
Hi @7rulnik thanks for this suggestion! 

[...]

PR #727, opened 2023-01-18
bugfix(cache): ignores unfollowables in content cache strategy

sverweij:
## Description

The list of 'changes' contains files/  modules dependency-cruiser was not  able to 'follow' (e.g.  because it's an unsupported file type) with changeType _deleted_  - e.g. css or  picture  imports in react projects. Apart from that these entries are  superfluous and make the list larger than necessary, this caused the second run  without any changes to not use the cache.

> A beta  version of dependency-cruiser that includes the fix in this  PR has been published on npmjs:
> - `dependency-cruiser@12.5.1-beta-1`        
> - sha-sum: `c294e758a9dcd5828ebcdd2742c53fd7cd5948f3`

## Motivation and Context

Fixes a bug @7rulnik found  while validating #715 (thanks!)

## How Has This Been Tested?

- [x] green ci
- [x] additional unit test
- [x] run against facebook/react

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/src/cache/utl.js
+++ b/src/cache/utl.js
@@ -111,7 +111,10 @@ function moduleIsInterestingForDiff(pModule) {
     !pModule.consolidated &&
     !pModule.coreModule &&
     !pModule.couldNotResolve &&
-    !pModule.matchesDoNotFollow
+    !pModule.matchesDoNotFollow &&
+    // as followable is optional, !exists when the module _is_ followable
+    // explicit comparison with false
+    pModule.followable !== false
   );
 }
 
--- a/test/cache/find-content-changes.spec.mjs
+++ b/test/cache/find-content-changes.spec.mjs
@@ -73,6 +73,34 @@ describe("[U] cache/find-content-changes - cached vs new", () => {
     ]);
   });
 
+  it("returns files that have been earmarked as not followable as 'ignored'", () => {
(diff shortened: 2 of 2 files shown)


7rulnik:
@sverweij just tested it and it definitely works as expected. Thank you!

markmssd:
Hey! First of all, great work on getting the `content` cache strategy working, it's a massive performance improvement! ⚡

I'm facing an issue which I'm not sure if it's the expected behaviour. I execute the dependency-cruiser first, and it properly generates the cache (takes around 10 seconds). If I re-run it again without making any changes, it properly uses the cache and runs in about 1 second) 🎉

Now, if I modify any file, and re-run it, it completely ignores the cache and re-runs it from scratch again (~10 seconds). My expectation was that it would use the cache for all unmodified files, and re-run for the modified files only.

Let me know if that is the expected behaviour. If it is, I'd be glad to contribute and try to improve it of course!

sverweij:
Hi @markmssd you're right to expect that! However, it isn't yet implemented. I initially wanted to have that in (also for the metadata strategy) - but setting up the caching mechanism turned out to be challenge enough. It's still on the roadmap, and it's good to know there's interest!

PR #760, opened 2023-03-17
fix: Add support for typescript v5

lukeapage:
Fixes #579

### DIFF
--- a/src/meta.js
+++ b/src/meta.js
@@ -12,7 +12,7 @@ module.exports = {
     livescript: ">=1.0.0 <2.0.0",
     svelte: ">=3.0.0 <4.0.0",
     swc: ">=1.0.0 <2.0.0",
-    typescript: ">=2.0.0 <5.0.0",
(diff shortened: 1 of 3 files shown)


src/meta.js
```diff
@@ -12,7 +12,7 @@ module.exports = {
     livescript: ">=1.0.0 <2.0.0",
     svelte: ">=3.0.0 <4.0.0",
     swc: ">=1.0.0 <2.0.0",
-    typescript: ">=2.0.0 <5.0.0",
+    typescript: ">=2.0.0 <6.0.0",
```
sverweij, line 15: `src/meta.js` is generated with a script in `tools` (from `package.json`; meta.js contains some other information from the package manifest as well), so these changes will be overwritten unless they're added to `package.json` as well.

I've added a commit to the PR to that effect.

sverweij:
Hi @lukeapage thanks for the PR! I've made a small modification to prevent that when the release script runs your changes won't be overwritten.

PR #764, opened 2023-03-25
Replace "args" with "arguments" in config file template

greglockwood:
## Description

Changed the use of `args` to `arguments` in the config JS file template in the context of the `options.webpackConfig` section.

## Motivation and Context

Our webpack config exports a function that relies on the second parameter to retrieve the `mode` (e.g. `'production'`). So I customized the generated config file to set `options.webpackConfig.args` to `{ mode: 'production' }`. When I attempted to run `dependency-cruiser` using it, I got a `TypeError` about trying to access the `mode` property on a null value. When I logged out the values, I saw that the second value was indeed coming in as null. So I looked in the code to see where it was reading the `webpackConfig` settings and found that the correct property name was `arguments`, not `args`.

So I figured I would update the template that generates the config so that others don't get stuck as I did.

## How Has This Been Tested?

I haven't, to be honest, since it was simple text changes to a template file.

- [ ] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/.dependency-cruiser.json
+++ b/.dependency-cruiser.json
@@ -281,14 +281,14 @@
       The (optional) fileName attribute specifies which file to take (relative to dependency-cruiser's
       current working directory. When not provided defaults to './webpack.conf.js'.
 
-      The (optional) `env` and `args` attributes contain the parameters to be passed if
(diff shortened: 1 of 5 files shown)


sverweij:
Thanks @greglockwood!

PR #804, opened 2023-04-30
perf(extract): splits compiler availability and execution

sverweij:
## Description

- repurposes src/extract/transpile/meta.mjs as a main entry for transpilation & data _about_ transpilation options available in the environment to just the data about transpilation.
- adds a new utility to check whether a transpiler is available - one that just require.resolve's the thing instead of importing or requiring it. This is a lot faster, especially when checking for big transpiler modules (e.g. typescript, babel).
- moves transpilation itself to src/extract/transpile/index.mjs

## Motivation and Context

- Better design wise 
The meta.mjs module now only has one responsibility: providing information about the environment.

- Gives better performance
It reduces dependency-cruiser's startup time - loading the transpilers is now shifted to _after_ checking the cache, so 100% cached runs are quite a bit faster. For normal runs there is almost no difference, though. Yet.

- Enables some new performance optimizations.
Now there's only one entry point for the transpilers (and only one module using them), it's a lot easier to optimize that as well (will be in a separate PR or commit).

## How Has This Been Tested?

- [x] green ci
- [x] adapted automated non-regression tests

#### 100% served from cache
```
bin/dependency-cruise.mjs src bin test configs types tools --ignore-known --cache node_modules/.cache/dependency-cruiser/load-cached --no-progress
  Time (mean ± σ):      1.184 s ±  0.014 s    [User: 1.232 s, System: 0.191 s]
  Range (min … max):    1.169 s …  1.225 s    30 runs
```

#### without a cache

```
 bin/dependency-cruise.mjs src bin test configs types tools --ignore-known --no-cache --no-progress
  Time (mean ± σ):      3.074 s ±  0.028 s    [User: 4.757 s, System: 0.319 s]
  Range (min … max):    3.044 s …  3.203 s    30 runs
```

#### 100% served from cache

On my old, 4 core mac it's about 2.4x faster than before this PR:

```
Benchmark 1: bin/dependency-cruise.mjs src bin test configs types tools --ignore-known --cache node_modules/.cache/dependency-cruiser/load-cached --no-progress
  Time (mean ± σ):     490.1 ms ±   6.8 ms    [User: 520.5 ms, System: 113.4 ms]
  Range (min … max):   483.9 ms … 512.8 ms    30 runs
```

#### without a cache

About the same performance (the mean is 1.005x slower, but these 0.016 s are still within the σ of both runs).

```
bin/dependency-cruise.mjs src bin test configs types tools --ignore-known --no-cache --no-progress
  Time (mean ± σ):      3.090 s ±  0.026 s    [User: 4.804 s, System: 0.326 s]
  Range (min … max):    3.055 s …  3.176 s    30 runs
```

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


src/cli/index.mjs
```diff
@@ -60,7 +60,7 @@ async function extractTSConfigOptions(pCruiseOptions) {
     const { default: extractTSConfig } = await import(
       "../config-utl/extract-ts-config.mjs"
     );
-    lReturnValue = extractTSConfig(lTSConfigFileName);
+    lReturnValue = await extractTSConfig(lTSConfigFileName);
```
sverweij, line 63:
```suggestion
    lReturnValue = extractTSConfig(lTSConfigFileName);
```

Issue #121, opened 2019-04-14
Not working on TypeScript classes


Ivanca:
Tried to use the tool without much success

![image](https://user-images.githubusercontent.com/907138/56087727-3c095400-5e36-11e9-8089-e2a78f0b7de5.png)

It shows that little despite AppManager having dozens of imports

![image](https://user-images.githubusercontent.com/907138/56087735-6eb34c80-5e36-11e9-99f1-31fb13716101.png)

For reference the command I'm using is 
```
depcruise --max-depth 99 --output-type dot .\src\scripts\main.ts | dot -T svg > dependencygraph.svg
```

When I try to use it on the all folder the result is empty, meaning when I run:
```
depcruise --max-depth 99 --output-type dot src/scripts | dot -T svg > dependencygraph.svg
```

### Expected Behavior
It should show dependencies

### Current Behavior
It doesn't show dependencies

sverweij:
From the diagram it seems the AppManager module cannot be found. This can have several causes. Assuming that AppManager exists, and the import reference in `src/scripts/main` is correct it might be dependency-cruiser can't find the typescript compiler (it searches for it in the same spot it's installed). 

What does `depcruise --info` tell in your installation?

Ivanca:
Typescript shows with ✖ but typescript is already installed in node_modules? (I verified using npm list)

I tried installing dependency-cruiser locally to see if that would help but did nothing.

sverweij:
- If you've installed dependency-cruiser globally typescript needs to be installed globally as well in order for dependency-cruiser to find it. Judging from the command line script in your issue, you're using that.
- If you run dependency-cruiser from a local install (so it's in your projects' `node_modules` and the binary is in `./node_nodules/.bin/depcruise`) and typescript is installed locally, dependency-cruiser picks up that one.

Note that if you use `depcruise` on the command line you're using the globally installed one. So even if you install dependency-cruiser locally, it'll use the global one, and pick up any global compiler.

So you can do one of two things:
- install typescript globally as well if you use the global depcruise
- make sure to run the local depcruise script - either by calling call `./node_modules/.bin/depcruise` directly (it's where the _scripts_ section of your package.json will look) or use npm's npx : `npx depcruise ...`

HTH

sverweij:
Hi @Ivanca I assume the answer in the previous comment sufficed to resolve the issue. If that isn't the case, feel free to re-open the issue with more details.

I'll add an entry to the FAQ as this is bound to come up more often.

PR #815, opened 2023-06-16
Use transpilers installed locally first (regardless of install location)

ypresto:
## Description

With this PR, dependency-cruiser will find locally installed transpilers first, then fallback to original behavior. So depcruise will work out-of-the-box with local transpilers regardless of `npx -p` or `npm i -g`.

Depends on this PR: https://github.com/sverweij/semver-try-require/pull/14

## Motivation and Context

Current depcruise will find transpilers in where depcruise itself is installed.
It means, transpilers installed in node_modules of local project are not used, unless depcruise is also installed locally.

https://github.com/sverweij/dependency-cruiser/issues/121#issuecomment-483259499
https://github.com/sverweij/dependency-cruiser/pull/661

Not only this PR solves setup issues, even we does not need to add depcruise to devDependencies. I wanted this to keep the number of dependencies fewer (e.g. for CI).

## How Has This Been Tested?

```
cd /path/to/semver-try-require # modified version
npm pack
cd /path/to/dependency-cruiser
npm i /path/to/semver-try-require/*.tgz
```

### Test for `npx`

```
cd /path/to/dependency-cruiser
rm -rf node_modules/typescript node_modules/@swc
cd /path/to/typescript-project

# ok
npx -p /path/to/dependency-cruiser depcruise -i 
npx -p ~/github/dependency-cruiser depcruise -p cli-feedback src --output-type dot --include-only "..." --ts-pre-compilation-deps
rm -rf node_modules/typescript

# ng
npx -p /path/to/dependency-cruiser depcruise -i # ng
npx -p ~/github/dependency-cruiser depcruise -p cli-feedback src --output-type dot --include-only "..." --ts-pre-compilation-deps
```

### Test for global install

```
cd /path/to/dependency-cruiser
npm pack
npm i -g dependency-cruiser-13.0.3.tgz
# we need to install locally referenced dependency manually. in my case:
npm i -g semver-try-require

cd /path/to/typescript-project

# ok
depcruise -i 
depcruise -p cli-feedback src --output-type dot --include-only "..." --ts-pre-compilation-deps


rm -rf node_modules/typescript
# ng
depcruise -i 
depcruise -p cli-feedback src --output-type dot --include-only "..." --ts-pre-compilation-deps

npm i -g typescript
# ok
depcruise -i 
depcruise -p cli-feedback src --output-type dot --include-only "..." --ts-pre-compilation-deps
```

### Of course also works when locally installed

```
cd /path/to/typescript-project
npm i -D /path/to/dependency-cruiser/dependency-cruiser-13.0.3.tgz

# ok
depcruise -i 
depcruise -p cli-feedback src --output-type dot --include-only "..." --ts-pre-compilation-deps
```

- [ ] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -10,7 +10,7 @@
 
 ### Q: TypeScript, CoffeeScript, LiveScript or Vue Single File Component (SFC) dependencies don't show up. How can I fix that?
 
-**A**: Install the compiler you use in the same spot dependency-cruiser is installed (or vv).
+**A**: Install the compiler you use in your node_modules, or the same spot dependency-cruiser is installed (or vv).
 
 For some types of TypeScript dependencies you need to flip a switch,
 which is what the next question is about.
@@ -26,16 +26,18 @@ running `depcruise --info`.
 
 When it turns out they aren't yet:
 
-- if you're running dependency-cruiser as a local (development-)
-  dependency (**recommended!**), install the necessary transpilers there.
-- if you're running dependency-cruiser as a global install, install
-  the necessary transpilers globally as well.
-- if you're running dependency-cruiser with npx you can pass the necessary
-  compiler(s) with the `-p` option. E.g. to get a list of all incoming and
-  outgoing dependencies of a TypeScript file you could use this:
-  ```
-  npx -p typescript@4.3.5 -p dependency-cruiser@latest depcruise src -T text --focus src/main/index.ts
-  ```
+- if transpilers are installed in your project (the node_modules directory reachable
+  from the current directory), dependency-cruiser will find them automatically.
+- if you don't want to install transpilers as dependencies of your project,
+  follow the instructions below:
+  - if you're running dependency-cruiser as a global install (npm i -g), install
+    the necessary transpilers globally.
+  - if you're running dependency-cruiser with npx, you can pass the necessary
+    compiler(s) with the `-p` option. E.g. to get a list of all incoming and
+    outgoing dependencies of a TypeScript file you could use this:
+    ```
+    npx -p typescript@4.3.5 -p dependency-cruiser@latest depcruise src -T text --focus src/main/index.ts
+    ```
 
 </details>
 
--- a/package.json
+++ b/package.json
@@ -163,7 +163,6 @@
     "ignore": "5.2.4",
     "indent-string": "5.0.0",
     "interpret": "^3.1.1",
-    "is-installed-globally": "0.4.0",
     "json5": "2.2.3",
     "lodash": "4.17.21",
     "prompts": "2.4.2",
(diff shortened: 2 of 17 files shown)


ypresto:
Maybe we need --trust-foobar flag? ("Workspace Trust" in vscode)

sverweij:
Hi @ypresto thank you for this pull request! And sorry it took such a long time to come up with a response. 

The feature you're aiming for will be useful if you don't want to install dependency-cruiser as part of a project (yet) - I like the idea! We'll have to add some caveats (the version of globally installed compilers might differ from the local ones, compilers might behave slightly differently when run globally, generally it's a good idea to keep the used compilers close to the projects), but for quickly surveying a new codebase each of these caveats can be a useful trade-off.

On the implementation: 

First off: the PR shows you've taken the time to understand the codebase _and_ put in a non-trivial amount of time into the implementation - I very much appreciate that!

- semver-try-require (which nowadays should better be called semver-try-import ...) indeed seems the right 'choke-point' to implement this. 
- In v13 dependency-cruiser was migrated to ESM as much as possible to make it work  with current features of the node platform and to future proof it - the whole ecosystem is moving in that direction as well. This pull request re-introduces commonjs in spots that were just migrated to ESM. I will consider merging this PR if we can attain the same goal, but instead with ESM.
- I do understand that in its current state additional automated tests are not yet part of the pull request. We will need them, though, before we can merge the PR. The additional tests will both help understand the added code and (once merged) will prevent non-regression.

ypresto:
Thank you for reviewing this! :heart:

> semver-try-require indeed seems the right 'choke-point'

That's right, but its change is small that just passing a `path` argument to `createRequire()`.
https://github.com/sverweij/semver-try-require/pull/14/files#diff-31bcdc6df5f09515531a5c779db30d8281baea70f9adb2adefc07baa00b093e5R54-R57

(Off-topic):

On the other hand, there is similar but slightly different logics:

- `semver-try-require`
- `src/extract/transpile/vue-template-wrap.cjs`: with fallback module name
- `src/extract/transpile/try-import-available.mjs`: to get the version without actual import

Maybe we should move them all to the library to make sure the logic is consistent in every place.

> ESM as much as possible

Node.js has `import.meta.resolve(specifier, *parent*)` (`parent` argument seems non-standard feature) but it is behind an experimental flag and [is actively changed (v20)](https://nodejs.org/ar/blog/announcements/v20-release-announce).
So IMO it is preferable to stick with `createRequire()` until it becomes stable, leaving escape hatch to ESM.
(There is also a polyfill: https://www.npmjs.com/package/import-meta-resolve)

> additional automated tests

I'll work on this. 💪

ypresto:
Oh no, could you reopen this please...? :pray: @sverweij

PR #821, opened 2023-07-07
feature(report): adds azure-devops reporter

### LINKED ISSUE #820: Feature request: Formatter for Azure DevOps
Formatter specific to Azure DevOps which would allow to report warnings and errors in a native way (which would surface in the Azure DevOps UI).

### Context
We use Azure DevOps as our CI/CD tool and recently started using dependency-cruiser, which I personally think is amazing. Thank you for making it.

The challenge we have is that it's not easy to know what violations there are (if any) when looking at the build summary. 
In the "best" case scenario the build fails (in case of violations with error severity), the error surfaced to the UI says `bash exited with non-zero exit code`, and the user needs to go into logs of the specific task to know what the violations were.
In the "worst" case there are violations which have warning severity, the build passes, everything is green and the user might not even be aware that there were some warnings.

### Expected Behavior
Violations detected by dependency-cruiser are surfaced in the Azure DevOps UI natively.

### Current Behavior
Violations are buried inside the logs.

### Possible Solution
Similarly to the TeamCity format, it should be possible to offer Azure DevOps format which could leverage the ["log issue" task command](https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#task-commands) to surface errors and warnings.
Additionally the ["complete" task command](https://learn.microsoft.com/en-us/azure/devops/pipelines/scripts/logging-commands?view=azure-devops&tabs=bash#complete-finish-timeline) could be used to set the status of the task/job to "SucceededWithIssues" when there are only violations with warning severity and no errors.

FYI a formatter for eslint exists which leverages the above: [eslint-formatter-azure-devops](https://github.com/EngageSoftware/eslint-formatter-azure-devops)

### Considered alternatives
I'm currently implementing an alternative solution using [BuildQualityChecks Azure DevOps extension](https://marketplace.visualstudio.com/items?itemName=mspremier.BuildQualityChecks), specifically its [Warnings Policy](https://github.com/MicrosoftPremier/VstsExtensions/blob/master/BuildQualityChecks/en-US/WarningsPolicy.md).

It seems like it generally works, but:
* The documentation of this extension says it mainly works with issues reported natively, offering option to pickup warnings from logs based on regex is just an addition.
* When parsing the logs, we're limited in identifying issue categories, which file it originates from, line #, etc.
* The warnings picked up by parsing the logs are not counted towards the reported total number of warnings. This means we end up with confusing messages like `The number of selected warnings (57 out of 43) exceeds the threshold (50)!`. This can happen if there are 43 warnings reported natively (e.g. by eslint formatter) and additional 14 warnings originating from parsing the output of dependency-cruiser.
* We need to have another, custom, task/step in the pipeline on top of the extension itself to set the status of the whole build as "SucceededWithIssues" if warnings exist.

sverweij:
## Description

- adds  an  `azure-devops` reporter

TODO
- [x] documentation
- [x] on par with the err and teamcity reporters for the summary
- [x] on par with the err and teamcity reporters for the summary (while using ignored violations)
- [x] better on par with the err and teamcity reporters for the individual violations

## Motivation and Context

fixes #820 

## How Has This Been Tested?

- [x] green ci

## Screenshots

Current sample output of 

```
npm run depcruise --no-ignore-known -T azure-devops --no-progress
```

will yield
```
##vso[task.logissue type=error;sourcepath=src/stout.mjs]not-to-dev-dep: src/stout.mjs -> node_modules/chai/index.js
##vso[task.logissue type=error;sourcepath=src/graph-utl/consolidate-to-folder.mjs]not-unreachable-from-cli: src/graph-utl/consolidate-to-folder.mjs
##vso[task.logissue type=error;sourcepath=src/schema/baseline-violations.schema.mjs]not-unreachable-from-cli: src/schema/baseline-violations.schema.mjs
##vso[task.logissue type=error;sourcepath=src/stout.mjs]not-unreachable-from-cli: src/stout.mjs
##vso[task.logissue type=warning;sourcepath=src/stout.mjs]not-unreachable-from-test: src/stout.mjs
##vso[task.logissue type=warning;sourcepath=src/utl/array-util.mjs]utl-module-not-shared-enough: src/utl/array-util.mjs
##vso[task.logissue type=warning;sourcepath=src/utl/find-all-files.mjs]utl-module-not-shared-enough: src/utl/find-all-files.mjs
##vso[task.logissue type=warning;sourcepath=src/utl/wrap-and-indent.mjs]utl-module-not-shared-enough: src/utl/wrap-and-indent.mjs
##vso[task.complete result=Failed;]8 dependency violations (4 error, 4 warning/ informational). 526 modules, 1311 dependencies cruised
```

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


src/report/azure-devops.mjs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ *
+ * @param {import("../../types/violations.js").IViolation} pViolation
+ */
+function formatViolation(pViolation) {
+  return `##vso[task.logissue type=${formatSeverity(
+    pViolation.rule.severity
+  )};sourcepath=${pViolation.from}]${pViolation.rule.name}`;
```
jmarek-sky, line 48: I wonder if it would be useful to add `pViolation.to` to the message alongside `pViolation.rule.name`.
It could help understanding the details when there are multiple violations of the same rule originating from one file.

src/report/azure-devops.mjs
```diff
@@ -0,0 +1,103 @@
...
+}
+
+/**
+ *
+ * @param {number} pNumberOfErrors
+ * @returns
+ */
+function formatSuccess(pNumberOfErrors) {
+  return pNumberOfErrors === 0 ? "Succeeded" : "Failed";
```
jmarek-sky, line 57: Ideally it would return `"SucceededWithIssues"` if there were any warnings without errors.

src/report/azure-devops.mjs
sverweij: Yes it is - it's on the todolist.

src/report/azure-devops.mjs
sverweij: Ah ok - that's what the SucceededWithIssues is for. Noted; can be added.

sverweij: Added. Will be in the next beta.

src/report/azure-devops.mjs
sverweij: ...also for circular dependencies: the whole cycle, for 'via' rules the path found etc - it should be on-par with the `err` reporter.

sverweij: - [x] added those as well (see added commits) => will be part of the next beta

sverweij:
@jmarek-sky I now notice it's possible to put the rule name in the `code` field 

So instead of doing this:

```
##vso[task.logissue type=error;sourcepath=src/index.js]not-to-dev-dep src/index.js -> node_modules/eslint/lib/api.js
```

we can do this:

```
##vso[task.logissue type=error;sourcepath=src/index.js;code=not-to-dev-dep;]src/index.js -> node_modules/eslint/lib/api.js
```

... which also seems what [eslint-formatter-azure-devops](https://github.com/EngageSoftware/eslint-formatter-azure-devops/blob/master/index.js#L33) is doing.

I'll make that modification later in the weekend & publish that as dependency-cruiser@13.1.0-beta-3 .

src/report/azure-devops.mjs
```diff
@@ -0,0 +1,222 @@
...
+  pNumberOfErrors,
+  pNumberOfWarns,
+  pNumberOfInfos,
+  pNumberOfIgnored
+) {
+  if (pNumberOfErrors > 0) {
+    return "Failed";
+  }
+  if (pNumberOfWarns + pNumberOfInfos + pNumberOfIgnored > 0) {
+    return "SucceededWithIssues";
```
jmarek-sky, line 113: On second thought, I'm no longer sure we should be using this status here.
Reasoning:
* it will change the status of the entire build even if there's just one warning
* the user has no way of controlling that

I wonder if it wouldn't be "nicer" to let the users decide what to do in some way.
One idea that comes to my mind is to not use this status here and for user to use the [BuildQualityChecks](https://marketplace.visualstudio.com/items?itemName=mspremier.BuildQualityChecks) extension to implement custom quality rules e.g. "set pipeline status to 'SucceededWithIssues' when warnings are > 20" and "fail pipeline when warnings are > 50". This might be more flexible and adaptable to various peoples' needs.

sverweij: Added in 47e143

> It might be nice to add a --max-warnings & --max-ignored to dependency-cruiser, just like [eslint's --max-warnings](https://eslint.org/docs/latest/use/command-line-interface#--max-warnings) 🤔 . Future feature, though.

PR #827, opened 2023-08-05
refactor(test): replaces mocha with node:test

sverweij:
## Description

- replaces mocha with node:test

### This won't be merged
Everything works. Had to write a little test reporter to get output that is to the point (no thousands of tests over the screen, but just a _little_ bit of progress information and only full information if a test fails). 

However, this PR is not going to be merged. I will continue to work with node:test projects that have (far) less automated tests as having no external dependencies for tests is still pretty neat. For dependency-cruiser, at this time, the drawbacks are heavier than the benefits.

### Why this is not going to be merged 1/2: node:test is too slow

There's a big drawback on using node:test on this amount of tests, though - node:test is very CPU hungry and _twice_ as slow as exactly the same tests run with `mocha`. This is not a problem when there's a moderate number of tests, but with ~1400 of them it really adds up.

The unix `time` utility emits numbers like this for node:test on my local machine:
```
yarn test  87,57s user 10,06s system 357% cpu 27,294 total
```

With mocha the same run looks like this:
```
yarn test  16,75s user 1,41s system 127% cpu 14,217 total
```

On the ci the slowdowns are comparable - see the ci runs associated with this PR.

### Why this is not going to be merged 2/2: challenges in filtering

One of the things a test runner should be able to do is quickly isolate one test, or a group of tests either by name (/ regex) or by filename. In addition this should be pretty fast

- `node --test` has a `--test-name-pattern` command line switch that does this, however
- it filters for the test name _only_, not for the test name + the suite. This means it's not easy to run just the tests in a suite. Worse, if your test names within a suite are somewhat generic (`it('returns a truthy value by default')` and appear in multiple places
- it's still quite slow even when targeting exactly one test - it looks like even skipping over tests takes it a while.

## Motivation and Context

reduces dependencies

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


jaads:
nice try! 👍 too bad that it didn't worked out. it seems from a certain project size on, third party libraries are still very  much needed.

PR #831, opened 2023-08-17
Suggest `typings` as a `mainFields` wherever `types` is suggested

BRMatt:
First off, I just want to say thanks very much for producing and maintaining this tool. I've been amazed at how fast/powerful it is, and the thorough documentation in config and READMEs has been super helpful! Thank you! ❤️ 🙇 

## Description

The FAQs mention that if types from a package cannot be resolved then the user should configure `mainFields` to include the keys that the package is using in its `package.json` file.

The generated config and documentation suggest using `main` and `types` as `mainFields`. However, [TypeScript documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html) notes that:

> ... the "typings" field is synonymous with types, and could be used as well.

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

Rather than make users of this package search through documentation and `node_modules` to discover this issue, this commit changes `dependency-cruiser` to support (or at least, suggest) both of these TypeScript defaults.

## Motivation and Context

One of our dependencies changed how they export types. We spent several confusing hours trying to track down the issue. It was especially confusing because we'd already configured dependency-cruiser to use `["main", "types"]` for `mainFields`, as the documentation suggests.

We were not aware that TypeScript also supported `typings` as a synonym for `types`. It would've saved us a lot of head scratching if `dependency-cruiser` had suggested both these values from the beginning.

## How Has This Been Tested?

- [x] green ci

## Types of changes

- [X] Documentation only change

Marking as a documentation change as it only affects documentation/the initial config that is generated by the tool.

## Checklist

- [X] :book:

- [X] :balance_scale:

### DIFF
--- a/doc/faq.md
+++ b/doc/faq.md
@@ -304,7 +304,7 @@ in that external package package.json e.g. like so:
   // rules: [],
   options: {
     enhancedResolveOptions: {
-      mainFields: ["main", "types"]
+      mainFields: ["main", "types", "typings"]
     }
     // other options ...
   }
--- a/src/cli/init-config/build-config.mjs
+++ b/src/cli/init-config/build-config.mjs
@@ -99,7 +99,7 @@ function buildWebpackConfigAttribute(pInitOptions) {
       // env: {},
       // arguments: {}
     },`
-    : `// webpackConfig: { 
+    : `// webpackConfig: {
     //  fileName: 'webpack.config.js',
     //  env: {},
     //  arguments: {}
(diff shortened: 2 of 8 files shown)


sverweij:
Hi @BRMatt thanks for the research and this clearly documented pull request 🤩 - future users will probably at least as thankful for this addition as I am.

I recalled two other places the mainFields attribute is documented and updated the documentation there as well.

The merged PR will be part of the next release (which I expect to take  place in the upcoming weekend).

BRMatt:
Thanks @sverweij!

PR #840, opened 2023-09-18
Added missing ": true," in options-reference.md example.

LartyHD:
The title or at least the on line code edit should say everything already.


sverweij:
Vielen dank für diesen Beitrag @LartyHD!

PR #875, opened 2023-11-24
feat(extract): detect 'type-only' with inline imports/exports

### LINKED ISSUE #873: Feature request: Consider TS `imports` with all `type` the same as `type-only`.
### Context

[TypeScript 4.5 added inline type imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names). `dependency-cruiser` has support for `'type-only'` in `dependencyTypesNot`, but this only works when using separate type imports, and not when all inline imports are qualified as `type`.

This can be fixed by moving the individual type imports to a separate type import, but this is cumbersome for some of us since using [ESLint's consistent-type-imports with `fixStyle: 'inline-type-imports'`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md#fixstyle) will prefer adding `type` inline even when all imports are `type`.

Some of us prefer this syntax since it makes it easier to mix type and non-type imports.

### Expected Behavior

This rule:

```js
{
  name: 'not-to-dev-dep',
  severity: 'error',
  comment:
    "This module depends on an npm package from the 'devDependencies' section of your " +
    'package.json. It looks like something that ships to production, though. To prevent problems ' +
    "with npm packages that aren't there on production declare it (only!) in the 'dependencies'" +
    'section of your package.json. If this module is development only - add it to the ' +
    'from.pathNot re of the not-to-dev-dep rule in the dependency-cruiser configuration',
  from: {
    path: '^(app)',
    pathNot: [
      '.(spec|test).(js|mjs|cjs|ts|ls|coffee|litcoffee|coffee.md)$',
      // allow dev depencendies in modules/testing as those are helpers for tests
      'modules/testing',
    ],
  },
  to: {
    dependencyTypes: ['npm-dev'],
    dependencyTypesNot: ['type-only'],
  },
},
```

Should not be fired when using imports like this:

```ts
import { type SomeType, type OtherType } from 'my-dev-dependency';
```

### Current Behavior

Correctly ignores `devDependencies` imports when using this syntax:

```ts
import type { SomeType, OtherType } from 'my-dev-dependency';
```

But this syntax trips the rule:

```ts
import { type SomeType, type OtherType } from 'my-dev-dependency';
```

### Possible Solution

All members inside the import statement are qualified as `type` so it could be considered as `'type-only'` for rule matching purposes.

### Considered alternatives

Possibly leave it as is and just force using `import type { ... }` explicitly. See "Context" section above on why it is inconvenient.

alvaro-cuesta:
## Description

Detect dependencies extracted from TypeScript as `'type-only'` also when all of the inline imports/re-exports are qualified as `type`.

## Motivation and Context

[TypeScript 4.5 added inline type imports](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html#type-modifiers-on-import-names). `dependency-cruiser` has support for `'type-only'` in `dependencyTypesNot`, but this only works when using type import statements, and not when regular import statements have all their inline imports qualified as `type`.

This can be fixed by moving the individual type imports to a separate type import statement, but this is cumbersome for some of us since using [ESLint's consistent-type-imports with `fixStyle: 'inline-type-imports'`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/consistent-type-imports.md#fixstyle) will prefer adding `type` inline even when all imports are `type`.

By extension this PR also implements the functionality for inline type re-exports.

Closes #873.

## How Has This Been Tested?

- [x] green ci
- [x] additional unit tests

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/src/extract/ast-extractors/extract-typescript-deps.mjs
+++ b/src/extract/ast-extractors/extract-typescript-deps.mjs
@@ -15,7 +15,7 @@ function isTypeOnlyImport(pStatement) {
       (pStatement.importClause.namedBindings &&
         pStatement.importClause.namedBindings.elements &&
         pStatement.importClause.namedBindings.elements.every(
-          (pE) => pE.isTypeOnly,
+          (pElement) => pElement.isTypeOnly,
         )))
   );
 }
@@ -32,7 +32,7 @@ function isTypeOnlyExport(pStatement) {
     // named reexports are per-element though
     (pStatement.exportClause &&
       pStatement.exportClause.elements &&
-      pStatement.exportClause.elements.every((pE) => pE.isTypeOnly))
+      pStatement.exportClause.elements.every((pElement) => pElement.isTypeOnly))
   );
 }
 
--- a/src/extract/ast-extractors/extract-typescript-deps.mjs
+++ b/src/extract/ast-extractors/extract-typescript-deps.mjs
@@ -8,16 +8,31 @@ const typescript = await tryImport(
   meta.supportedTranspilers.typescript,
 );
 
-function isTypeOnly(pStatement) {
+function isTypeOnlyImport(pStatement) {
   return (
-    (pStatement.importClause && pStatement.importClause.isTypeOnly) ||
+    pStatement.importClause &&
+    (pStatement.importClause.isTypeOnly ||
+      (pStatement.importClause.namedBindings &&
+        pStatement.importClause.namedBindings.elements &&
+        pStatement.importClause.namedBindings.elements.every(
+          (pE) => pE.isTypeOnly,
+        )))
+  );
+}
+
+function isTypeOnlyExport(pStatement) {
+  return (
     // for some reason the isTypeOnly indicator is on _statement_ level
     // and not in exportClause as it is in the importClause ¯\_ (ツ)_/¯.
     // Also in the case of the omission of an alias the exportClause
     // is not there entirely. So regardless whether there is a
     // pStatement.exportClause or not, we can directly test for the
     // isTypeOnly attribute.
-    pStatement.isTypeOnly
+    pStatement.isTypeOnly ||
+    // named reexports are per-element though
+    (pStatement.exportClause &&
+      pStatement.exportClause.elements &&
+      pStatement.exportClause.elements.every((pE) => pE.isTypeOnly))
   );
 }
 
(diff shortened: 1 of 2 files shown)


alvaro-cuesta:
Unsure if this should be considered a new feature or a fix, a bit of a middle ground (opted for feature because it feels like adding functionality to detect more combinations).

Also unsure if this should be considered a breaking change. I think there could be a custom rule that might start rejecting with this change, right? In particular if `'type-only'` is used for `dependencyTypes` instead of `dependencyTypesNot`, this were false negatives that the rule will start catching.

src/extract/ast-extractors/extract-typescript-deps.mjs
```diff
@@ -8,16 +8,31 @@ const typescript = await tryImport(
...
 
-function isTypeOnly(pStatement) {
+function isTypeOnlyImport(pStatement) {
+  return (
+    pStatement.importClause &&
+    (pStatement.importClause.isTypeOnly ||
+      (pStatement.importClause.namedBindings &&
+        pStatement.importClause.namedBindings.elements &&
+        pStatement.importClause.namedBindings.elements.every(
+          (pE) => pE.isTypeOnly,
```
sverweij, line 18:
```suggestion
          (pElement) => pElement.isTypeOnly,
```

src/extract/ast-extractors/extract-typescript-deps.mjs
```diff
@@ -8,16 +8,31 @@ const typescript = await tryImport(
...
     // Also in the case of the omission of an alias the exportClause
     // is not there entirely. So regardless whether there is a
     // pStatement.exportClause or not, we can directly test for the
     // isTypeOnly attribute.
-    pStatement.isTypeOnly
+    pStatement.isTypeOnly ||
+    // named reexports are per-element though
+    (pStatement.exportClause &&
+      pStatement.exportClause.elements &&
+      pStatement.exportClause.elements.every((pE) => pE.isTypeOnly))
```
sverweij, line 35:
```suggestion
      pStatement.exportClause.elements.every((pElement) => pElement.isTypeOnly))
```

sverweij (approved):
@alvaro-cuesta that was _very_ quick! And a great PR 

- good documentation, 
- thorough unit tests, 
- comments where it adds value to future maintainers and 
- a refactor of the isTypeOnly function that makes it more readable

Thanks!

sverweij:
> Unsure if this should be considered a new feature or a fix, a bit of a middle ground (opted for feature because it feels like adding functionality to detect more combinations).

Agree.

> Also unsure if this should be considered a breaking change (opted for "yes" but I'd appreciate your perspective): I think it's possible that there's a rule out there that might start rejecting with this change, right? My thought process was: `dependencyTypesNot` is less problematic since it will just start accepting more stuff, but if someone used `'type-only'` in a rule using `dependencyTypes`, it might have been ignoring false negatives that the rule will start catching after this change.

Yep - it's the bane of maintaining a code analysis/ linting packages that improvements sometimes are breaking changes, even though the number of affected consumers might be close to zero. The other open PR (#867) has the same issue.

PR #950, opened 2024-07-20
refactor: replaces lodash

sverweij:
## Description

- [x] centralizes remaining uses of lodash in central utilities
- [x] in the central utilities replace the lodash array functions with native alternatives that work in our context
- [x] in the central utilities replace the lodash object functions with native alternatives that work in our context
- [x] removes lodash from the package manifest

## Motivation and Context

lodash is a great library, however 
- it's not really maintained anymore
- it has fairly large download size (1.4Mb [according to pkg-size](https://pkg-size.dev/lodash)) - while we only use a small subset of it

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


src/config-utl/extract-depcruise-config/merge-configs.mjs
```diff
@@ -1,6 +1,6 @@
 import { isDeepStrictEqual } from "node:util";
-import uniqBy from "lodash/uniqBy.js";
-import uniqWith from "lodash/uniqWith.js";
+
+import { uniqBy, uniqWith } from "#utl/array-util.mjs";
```
sverweij, line 3:
```suggestion
import { uniqBy, uniqWith } from "#utl/array-util.mjs";
```

src/main/rule-set/assert-validity.mjs
```diff
@@ -1,11 +1,11 @@
 import Ajv from "ajv";
 import safeRegex from "safe-regex";
-import has from "lodash/has.js";
-import get from "lodash/get.js";
 import { assertCruiseOptionsValid } from "../options/assert-validity.mjs";
 import { normalizeToREAsString } from "../helpers.mjs";
 import configurationSchema from "#configuration-schema";
 
+import { has, get } from "#utl/object-util.mjs";
```
sverweij, line 7:
```suggestion
import { has, get } from "#utl/object-util.mjs";
```

src/main/rule-set/assert-validity.mjs
```diff
@@ -29,7 +29,7 @@ function assertSchemaCompliance(pSchema, pConfiguration) {
 }
 
 function hasPath(pObject, pSection, pCondition) {
-  // pCondition can be nested properties, so we use lodash.has instead
+  // pCondition can be nested properties, so we use _.has instead
   // of elvis operators
```
sverweij, line 33:
```suggestion
  // pCondition can be nested properties, so we use a bespoke
  // 'has' function instead of simple elvis operators
```

src/report/dot/index.mjs
```diff
@@ -151,7 +151,7 @@ function pryReporterOptionsFromResults(pGranularity, pResults) {
   const lFallbackReporterOptions =
     pResults?.summary?.optionsUsed?.reporterOptions?.dot;
 
-  // using lodash.get here because the reporter options will contain nested
+  // using _.get here because the reporter options will contain nested
   // properties, which it handles for us
```
sverweij, line 155:
```suggestion
  // using a bespoke 'get' function here because the reporter options will
  // contain nested properties, which it handles for us
```

src/report/dot/theming.mjs
```diff
@@ -14,10 +13,14 @@ function matchesCriterion(pModuleKey, pCriterion) {
...
   return Object.keys(pSchemeEntry.criteria).every((pKey) => {
-    // we use lodash.get here because in the criteria you can enter
-    // nested keys like "rules[0].severity" : "error", and lodash.get handles
+    // the keys can have paths in them like {"rules[0].severity": "info"}
+    // To get the criterion treat that key as a string and not as a path
+    // eslint-disable-next-line security/detect-object-injection
+    const lCriterion = pSchemeEntry.criteria[pKey];
+    // we use _.get here because in the criteria you can enter
+    // nested keys like "rules[0].severity" : "error", and _.get handles
     // that for us
```
sverweij, line 22:
```suggestion
    // we use a bespoke 'get' here because in the criteria you can enter
    // nested keys like "rules[0].severity" : "error", and that function
    // handles those for us
```

src/report/dot/theming.mjs
```diff
@@ -14,10 +13,14 @@ function matchesCriterion(pModuleKey, pCriterion) {
...
-    const lCriterion = get(pSchemeEntry.criteria, pKey);
+    // console.error(pSchemeEntry.criteria, pKey, pc.bold(lCriterion));
```
sverweij, line 23:
```suggestion
```

test/cli/index.spec.mjs
```diff
@@ -203,6 +203,7 @@ function setModuleType(pTestPairs, pModuleType) {
 
 function runFileBasedTests(pModuleType) {
   setModuleType(TEST_PAIRS, pModuleType).forEach((pPair) => {
+    // if (pPair.description === "regular dot with default theme")
```
sverweij, line 206:
```suggestion
```

Issue #330, opened 2020-07-06
Add a way to match a rule based on the relative path from "from" to "to"


forty:
I would like to be able to write rules such as

```
    {
        name: 'no-parent',
        severity: 'error',
        comment: 'cannot require from parent directory',
        from: {
            path: '^src/' 
        },
        to: {
           relativePath: '^../'
        }
    }
```
and
```
    {
        name: 'no-inside',
        severity: 'error',
        comment: 'cannot require for than 1 level deep inside a folder',
        from: {
            path: '^src/' 
        },
        to: {
           relativePathNot: '^(\.\.?/)([^/]+)$|^(\.\.?/)([^/]+)/index.ts$'
        }
    }
```
The rule would match if `path.relative(path.dirname(from), to)` matches the regex (or doesn't match the regex for the relativePathNot)

### Context

I'm trying to do something similar to the "no-inside" rule above. Basically want to consider all folder as module boundaries. So for example:
`A.ts` can require `./B.ts` (or `./B/index.ts`), `../B.ts`, `../../../B.ts` but NOT `./b/B.ts` or `../b/B.ts`

I tried to do my best with regexp, but I'm not really sure it's possible to simulate that with regex alone :)

### Expected Behavior

Be able to match based on relative paths

### Current Behavior

Only absolute path can be user

### Possible Solution

Add this new matcher property as suggested above

### Considered alternatives

many complex regexp with all the possible depth of folder....

jomi-se:
Would be useful to have this. Maybe a way to ease the implementation would be to add an extra property such as `relativeTo: true`

sverweij:
Hi @forty thanks for this suggestion. Could [capture groups/ group matching](https://github.com/sverweij/dependency-cruiser/blob/develop/doc/rules-reference.md#group-matching) fit your use cases? With this you have the from's path or any part of it you need to at your disposal. Below is an example 

I'm reasonably confident the end result you had in mind is achievable with capture groups as well - but let me know if you need help with it!

So how does it work? You can but brackets around parts of the regular expression in the from part of the rule. The substring matches are available for later use in the `to` part of the rule with `$n` variables.  E.g. this rule makes sure modules only require from the same folder or deeper:

```javascript
{
  "allowed": [
    {
      "from": { 
        "path": "(.+/)[^/]+$" // will match e.g. src/cool/stuff/index.ts
      },
      "to": {
[...]

PR #961, opened 2024-10-27
feat: adds 'ancestor' matcher

sverweij:
## Description

- adds an `ancestor` matcher

## Motivation and Context

It's _probably_ possible to do this with regular expressions and group matching, but so far I haven't found a simple way to do that.

## How Has This Been Tested?

- [x] green ci
- [x] additional automated integration tests
- [x] additional automated unit tests
- [x] self-run (added a no-ancestors rule to our own .dependency-cruiser.mjs)

## Screenshots

You can now do things like this:

```javascript
// in the `forbidden` section of a dependency-cruiser config:
{
  name: "no-ancestor",
  comment: "don't allow importing from folders above the current one",
  severity: "error",
  from: {},
  to: { "ancestor": true }
}
```

Which, when run as part of a validation, would yield errors on every dependency that goes folders _up_ in stead of down:

sample-ancestor

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


1000i100:
Isn't it ready to merge ?

sverweij:
> Isn't it ready to merge ?

It is. I have doubts about the feature though. It does address an immediate ask, but is it sufficiently flexible for slightly more advanced (but just as valid in their use cases) rules regarding parent-child (e.g. relative paths https://github.com/sverweij/dependency-cruiser/issues/330) )? 

> I usually need an afternoon of quiet time to think this through, but in the past year such afternoons weren't possible. Another option is to just yolo-merge it, and live with the consequences of supporting it ...

sverweij:
Yolo & learn

PR #972, opened 2024-12-05
Update README.md

YozhEzhi:
Typo

depcruise  -> dependency-cruiser

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ pnpm add -D dependency-cruiser
 ### ... and generate a config
 
 ```shell
-npx depcruise --init
+npx dependency-cruiser --init
 ```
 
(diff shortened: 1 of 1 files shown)


sverweij:
Hi @YozhEzhi thanks for the PR. `depcruise` is a valid command (once dependency-cruiser is installed) - it's defined in the [package manifest](https://github.com/sverweij/dependency-cruiser/blob/main/package.json#L77). I've used it in the README (and other spots throughout the documentation) since ~2016 as it is the shorter variant of those available to invoke dependency-cruiser (`dependency-cruiser` is another), so it's not really a typo.

Is there another reason to prefer using the longer variant in the documentation?

PR #983, opened 2025-02-09
feat(extract): improves detection of protocol-only builtins (node & bun)

sverweij:
## Description

- uses `modules.isBuiltin` instead of the `modules.builtinModules` array, so throwing in builtin modules wih protocol prefixes doesn't need any pre- or post processing anymore. We can do this since now all versions of node we support have the `modules.isBuiltin` method available.
- adds `bun:` to the list of protocols we recognize.
- leaves the protocol with the module name for builtins that are only available with the protocol in the name (e.g. node:test, node:sqlite, node:sea or bun:ffi).

- [x] To decide - previously if you had both `fs` and `node:fs` in your source code, both would map to `fs`, nicely de-duplicating them. 
	- [x] Should we bring that back somehow? Yes
	- [x] If we don't are we going to run into backwards compatibility issues (and how serious might these be)? We will, not sure how much
	- [x] Can we mitigate? Yes - implemented in this PR

## Motivation and Context

- Makes it easier to support new builtin modules that follow the new trend of being protocol-prefixed only.
- Now out of the box supports recognizing node:sqlite, node:sea, node:test/reporters as core modules.
- Adds support for recognizing modules prototype-prefixed with `bun:` (which is relevant if you use dependency-cruiser with bun).

## How Has This Been Tested?

- [x] green ci
- [x] updated and additional automated non-regression tests

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/src/extract/helpers.mjs
+++ b/src/extract/helpers.mjs
@@ -46,11 +46,13 @@ export function extractModuleAttributes(pString) {
   let lReturnValue = { module: pString };
   const lModuleAttributes = pString.match(
     // eslint-disable-next-line security/detect-unsafe-regex
-    /^(?<protocol>node:|file:|data:)(?:(?<mimeType>[^,]+),)?(?<module>.+)$/,
+    /^(?<protocol>node:|file:|data:|bun:)(?:(?<mimeType>[^,]+),)?(?<module>.+)$/,
   );
 
   if (lModuleAttributes?.groups) {
-    lReturnValue.module = lModuleAttributes?.groups.module;
+    lReturnValue.module =
+      String(lModuleAttributes?.groups?.protocol ?? "") +
+      lModuleAttributes?.groups.module;
     if (lModuleAttributes?.groups?.protocol) {
       lReturnValue.protocol = lModuleAttributes.groups.protocol;
     }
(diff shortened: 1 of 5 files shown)


src/extract/helpers.mjs
```diff
@@ -28,6 +28,37 @@ export function detectPreCompilationNess(pTSDependencies, pJSDependencies) {
...
+
+/**
+ * Given a module name returns the canonical module name.
+ * @param {string} pProtocol
+ * @param {string} pModuleName
+ * @returns {string}
+ */
+function getCanonicalModuleName(pModuleName, pProtocol) {
+  const lModuleWithProtocol = pProtocol + pModuleName;
+  const lIsJsIshModule = pProtocol === "node:" || pProtocol === "bun:";
```
sverweij, line 53:
```suggestion
  const lIsJsIshProtocol = pProtocol === "node:" || pProtocol === "bun:";
```

src/extract/helpers.mjs
```diff
@@ -28,6 +28,37 @@ export function detectPreCompilationNess(pTSDependencies, pJSDependencies) {
...
+  const lIsProtocolOnlyModule = PROTOCOL_ONLY_BUILTINS.has(lModuleWithProtocol);
+
+  if (!lIsJsIshModule || lIsProtocolOnlyModule) {
```
sverweij, line 56:
```suggestion
  if (!lIsJsIshProtocol || lIsProtocolOnlyModule) {
```

PR #1014, opened 2025-10-13
feat: add bun module support

1000i100:
## Description

1. Schema validation - Add "bun:" protocol to allowed protocols
2. Automatic detection - Modules with bun:* prefix are automatically recognized as "core" modules
3. Documentation URLs - Bun core modules link to https://bun.sh/docs/api/{module} in reports
4. Tests - Add 2 new tests (1 for isBuiltin, 1 for getURLForModule)

## Motivation and Context

I use bun and note node on my project. When I tryed dependency-cruiser, it throw me errors. Here is why.

## How Has This Been Tested?

I have added unit-test like named in description, see files diff for details
On my LMDE6 laptop, all test succeed.

- [x] green ci

## Types of changes
Not sure if you see it like feat or bugfix.

## Checklist

- [ ] :book:
I think it don't need, if you think it is, let me know.

- [x] :balance_scale:
  - The contribution is my own original work. (with help of Claude AI, yep)


sverweij:
Thanks for this one as well @1000i100 - it adds functionality without complicating the mental load for those who use it. When I implemented support for bun I didn't realize that `bun:` prefixed modules could already be unambiguously considered core modules; thanks for that insight & implementation!

sverweij:
Thanks for the rebase @1000i100

PR #1020, opened 2025-10-28
fix: call `ITranspilerWrapper.isAvailable()` correctly in svelte pre process

sushichan044:
> Leave an [issue](https://github.com/sverweij/dependency-cruiser/issues/new/choose) with a
> do that, so we can have a conversation before you start. This will save you time!

## Description

This PR fixes the incorrect usage of `ITranspilerWrapper.isAvailable()` in the Svelte preprocessing module.
The method was being accessed as a property instead of being called as a function, which prevented proper availability checks for transpilers.

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


sverweij:
Good catch @sushichan044, thanks!

sverweij:
FYI - released as part of https://github.com/sverweij/dependency-cruiser/releases/tag/v17.2.0

PR #1061, opened 2026-05-30
refactor(resolve)!: replaces tsconfig-paths-webpack-plugin with enhanced-resolve's own tsconfig-paths feature SLIGHTLY BREAKING

sverweij:
## Description

- replaces tsconfig-paths-webpack-plugin with enhanced-resolve's own tsconfig-paths feature

> It seems that in one edge-case enhanced-resolve reacts differently. References to config-paths defined as below (with no `*` on the LHS, but _with_ a `*` on the RHS don't resolve, where they _did_ in tsconfig-paths-webpack-plugin.
> TBD:
>  - is this pattern 'allowed'/ supported by the official typescript implementation?
>    -> as it turns out it is not rejected, but it doesn't do anything either. It
>       is / was something specific in the tsconfig-paths package.
>  - if so, maybe make a PR at enhanced-resolve. If not, maybe plonk in a workaround locally - or put it in the list of 'breaking changes' (we're major-bumping anyway).
>     -> breaking change it is.
> 
> ```js
> // ...
>     "paths": {
>       "@sister": ["./libs/zus/*"]
>     },
> // ...
> ```

## Motivation and Context

- With enhanced-resolve's 'native' support for tsconfig-paths it's not necessary anymore to use a separate plugin for it.
- tsconfig-paths-webpack-plugin currently seems to be nominally maintained anymore
- as a bonus we loose 1 direct and 9 indirect dependencies we don't have to worry about anymore.

## How Has This Been Tested?

- [x] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:

### DIFF
--- a/.c8rc.json
+++ b/.c8rc.json
@@ -1,9 +1,9 @@
 {
   "checkCoverage": true,
-  "statements": 99.96,
+  "statements": 99.97,
   "branches": 98.7,
   "functions": 100,
-  "lines": 99.96,
+  "lines": 99.97,
   "exclude": [
     "bin",
     "configs",
--- a/package-lock.json
+++ b/package-lock.json
@@ -25,7 +25,6 @@
         "rechoir": "0.8.0",
         "safe-regex": "2.1.1",
         "semver": "7.8.1",
-        "tsconfig-paths-webpack-plugin": "4.2.0",
         "watskeburt": "5.0.3"
       },
       "bin": {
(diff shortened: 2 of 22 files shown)


sverweij:
@copilot the check-windows job fails on the test step (_npx mocha --invert --fgrep "#do-not-run-on-windows"_), whereas the same step on linux (and locally macos) succeeds. 

As an expert in differences between linux and windows - can you analyze the root-cause of this issue (and propose a fix)?

sverweij:
@copilot while that seems to have lessened the issue, a new issue has popped up in the test step on windows. Can you check why this is is going wrong and propose a fix?

PR #1093, opened 2026-09-14
fix(extract/swc): finds the swc Visitor when bun hands over the default export

### LINKED ISSUE #1092: Issue: with parser 'swc' under bun, extraction fails with 'getDependencies is not a function'
### Expected Behavior

With `options.parser: "swc"`, dependency-cruiser run by bun cruises the same dependencies as dependency-cruiser run by node.

### Current Behavior

Under bun, extraction fails for every module:

```
new SwcDependencyVisitor(pExoticRequireStrings).getDependencies is not a function. (In 'new SwcDependencyVisitor(pExoticRequireStrings).getDependencies(pSwcAST)', 'new SwcDependencyVisitor(pExoticRequireStrings).getDependencies' is undefined)
... in src/a.ts
```

### Possible Solution

`src/extract/swc/dependency-visitor.mjs` takes the visitor with `const { Visitor } = await tryImport("@swc/core/Visitor.js", ...)`, and `tryImport` returns a module's default export when it has one. `@swc/core/Visitor.js` is CommonJS: it sets `__esModule` and `exports.default = Visitor`.

- Node makes the whole `module.exports` the default export of a CommonJS module, so `{ Visitor }` finds the class.
- Bun honours `__esModule` and makes `exports.default` the default export. `tryImport` then returns the `Visitor` class itself, `{ Visitor }` is `undefined`, and the module falls back to its empty class.

Taking `.Visitor` when it is there, and the module itself when it is not, works under both. A pull request follows.

### Steps to Reproduce (for bugs)

1. Create a `package.json` with the devDependencies `dependency-cruiser@18.2.0` and `@swc/core@1.16.2`. `main` / 18.3.0 has the same code.
2. Add `src/a.ts` containing `import { b } from "./b.ts"; export const a = b;` and `src/b.ts` containing `export const b = 1;`.
3. Add `.dependency-cruiser.cjs` containing `module.exports = { forbidden: [], options: { parser: "swc" } };`.
4. Run `node node_modules/dependency-cruiser/bin/dependency-cruise.mjs src --config .dependency-cruiser.cjs`. It prints `✔ no dependency violations found (2 modules, 1 dependencies cruised)`.
5. Run `bun node_modules/dependency-cruiser/bin/dependency-cruise.mjs src --config .dependency-cruiser.cjs`. It fails with the error above.

`bun run depcruise` on a machine without node fails the same way, since bun then runs the node shebang itself.

### Context

We run dependency-cruiser in Bun projects on TypeScript 7. Its package no longer exposes the compiler API, so `swc` is the parser, and today that means installing node on machines that otherwise need only bun.

### Your Environment

- Version used: 18.2.0 (the code is unchanged on `main`)
- Node version: none; bun 1.4.2 (node 24.18.0 works)
- Operating System and version: macOS 26, arm64

kurovskyiii:
## Description

Takes swc's `Visitor` from whatever `tryImport` returns: the module under node, the class itself under bun.

## Motivation and Context

Fixes #1092.

`@swc/core/Visitor.js` is CommonJS that sets `__esModule`. Bun therefore makes `exports.default`, the `Visitor` class, its default export, and `tryImport` returns that class. Destructuring `{ Visitor }` from the class gives `undefined`, and the swc extractor quietly falls back to its empty class, so every module fails with `getDependencies is not a function`. Under node, the default export is the whole `module.exports`, and nothing changes.

## How Has This Been Tested?

- With the reproduction from #1092, the swc parser now gives the same result under bun 1.4.2 as under node 24.18.0: `✔ no dependency violations found (2 modules, 1 dependencies cruised)`.
- `test/extract/swc/*.spec.mjs` pass (35). The build, lint and depcruise parts of `npm run check` pass, and lint-staged's `oxlint` and `prettier` leave the file as it is.
- The full mocha run aborts on this machine (macOS arm64, node 24.18.0) with a panic inside `@swc/core`'s native binding (`parse.rs:237: failed to read program file`), in the swc cases of `test/extract/get-dependencies.cjs.spec.mjs`. It does so on `main` as well, so I could not get a full local run.

No test is added: the new branch is taken only under bun, which the suite does not run. I'm happy to add one if you see a way you'd like it done.

- [ ] green ci

## Types of changes

## Checklist

- [x] :book:

- [x] :balance_scale:


src/extract/swc/dependency-visitor.mjs
```diff
@@ -4,10 +4,11 @@ import tryImport from "#utl/try-import.mjs";
 import meta from "#meta.cjs";
 
 /** @type {import('@swc/core/Visitor')} */
-const { Visitor } = await tryImport(
+const lSwcVisitorModule = await tryImport(
   "@swc/core/Visitor.js",
   meta.supportedTranspilers.swc,
 );
+const Visitor = lSwcVisitorModule.Visitor ?? lSwcVisitorModule;
```
sverweij, line 14:
```suggestion
// under bun lSwcVisitorModule resolves to something different
// as under node.js - see https://github.com/sverweij/dependency-cruiser/issues/1092
// for details. In order to enable running dependency-cruiser on both:
const Visitor = lSwcVisitorModule.Visitor ?? lSwcVisitorModule;
```

sverweij:
Thanks for this PR @kurovskyiii (as well as the issue, that provided sufficient context)! 

I'll see if I can incorporate run on bun on the ci as well (and fix whatever breaks) - apparently its implementation isn't as a 1:1 replacement for node.js :-)

kurovskyiii:
> Thanks for this PR @kurovskyiii (as well as the issue, that provided sufficient context)!
> 
> I'll see if I can incorporate run on bun on the ci as well (and fix whatever breaks) - apparently its implementation isn't as a 1:1 replacement for node.js :-)

Thanks for the quick merge, and for taking Bun into CI. That's more than I hoped for 🙂

For what it's worth: in our project (~600 TypeScript modules, parser: "swc") this was the only difference we hit. With the fix, dependency-cruiser under Bun cruises the same graph as under Node. If you'd like someone to try a branch or a pre-release on Bun, ping me and I'll run it.

One unrelated thing I noticed while testing: on macOS arm64 with Node 24.18, the full mocha run on main aborts with a panic in @swc/core's native binding (parse.rs:237: failed to read program file), in the swc cases of test/extract/get-dependencies.cjs.spec.mjs. Happy to open an issue with the details if that's useful.

sverweij:
I noticed some differences with running the test suites under bun, but am not sure when bun is using its own engine or uses node.js under the hood. `bun test` -> uses bun, but also its own test engine so obviously mocha specifics (`before`, `after` triggers) fail. `bun run test` (or `LANG=en_US.UTF-8 NODE_OPTIONS=--no-warnings NO_COLOR=1 bun mocha`) runs fine (no panics, just test execution), but I suspect these actually use node.js (envs: bun 1.4.2, node 26.6.0 and node 24.18.0 - macOS arm64) as these runs also passed full colors on dependency-cruiser@18.3.0...


> One unrelated thing I noticed while testing: on macOS arm64 with Node 24.18, the full mocha run on main aborts with a panic in @swc/core's native binding (parse.rs:237: failed to read program file), in the swc cases of test/extract/get-dependencies.cjs.spec.mjs. Happy to open an issue with the details if that's useful.
I _have_ noticed incomplete installs after bumping swc (e.g. distribution binary node_modules/@swc/core-darwin-arm64 missing from lockfile as well as node_modules - something something cache probably) - but couldn't reproduce this locally (or on ci).

sverweij:
B.t.w. I didn't see a reason to delay publishing the fix to npmjs, so since a few seconds it's on there in dependency-cruiser@18.3.1

PR #1095, opened 2026-09-15
fix(extract): pick up imports inside module blocks in the tsc extractor

arpitjain099:
The three tsc extractors read `pAST.statements` directly, under the assumption stated right above them:

```js
/*
 * Both extractImport* assume the imports/ exports can only occur at
 * top level. AFAIK this the only place they're allowed, so we should
 * be good. Otherwise we'll need to walk the tree.
 */
```

There is one other place TypeScript allows them, and it is the one that matters for `.d.ts` files: inside a module block, which is what `declare module "some-package" { ... }` compiles to. `ModuleBlock` is not in `pAST.statements`, so nothing inside one is ever looked at.

Same four statements, once at the top level of a `.d.ts` and once wrapped in an ambient declaration. Both files are accepted by `tsc --noEmit --strict` with the three packages stubbed in `node_modules`:

```
$ node bin/dependency-cruiser.mjs --ts-pre-compilation-deps --ts-config tsconfig.json --output-type json shim.d.ts
deps: []

$ node bin/dependency-cruiser.mjs --ts-pre-compilation-deps --ts-config tsconfig.json --output-type json top.d.ts
deps: ["another-package","legacy-package","reexported-package"]
```

The dependencies are absent from the graph, not merely unreported, so a rule over those edges cannot fire and the module reads as having no dependencies at all. The shape this shows up in is the hand written `.d.ts` shim or an `@types` package, where wrapping the whole thing in `declare module` is the normal way to write it.

### The change

`getStatements` returns a node's statements with any module blocks flattened into them, recursing so a nested block is covered as well, and the three extractors take their input from it. A module declaration with no block, `declare module "some-package";`, has no statements to flatten and is untouched.

Scope worth noting, so this does not read as bigger than it is: module augmentation is unaffected, because TS2667 forbids imports directly inside one, and a relative specifier inside an ambient module is illegal per TS2439. What is reachable is package specifiers, which is exactly the shim case above.

### Verification

`test/extract/tsc/ambient-modules.spec.mjs` covers an import, a re-export, an import-equals, a type only import, a nested module block, and a module declaration without a block. Five of the six fail on `main`; the sixth is the no-block case, which is there so a change that starts descending into things it should not is caught.

`mocha --no-config test/extract/tsc/*.spec.mjs` is 76 passing with the change. I could not run the whole suite locally: `npm test` aborts partway through on `main` as well, with an swc binding panic (`failed to read program file` from `binding_core_node/src/parse.rs:237`) that has nothing to do with this change. `prettier --check` is clean on both touched files.


sverweij:
Hi @arpitjain099 thanks for the PR and for learning me something new :-) . 

The changes look kosher, the ci is green and  the unit tests look nice (they don't only test the code but also help explain the addition 👍 ). However I'm also a bit sleepy, which is when I tend to miss things, so I'll look again tomorrow before merging (and I'll run an LLM on the code to see if they have some 2c).

arpitjain099:
Thanks, and the Copilot note above was right, so I would hold off merging the previous commit.

`declare namespace A.B {}` is a `ModuleDeclaration` whose body is another `ModuleDeclaration`, and only the innermost one carries the `ModuleBlock`. I was testing the immediate body, so the qualified form fell through and its imports were still dropped:

```
plain namespace          ["pkg-a"]
qualified namespace      []
qualified deeper         []
declare module string    ["pkg-m"]
```

044346e walks down to the innermost declaration, which covers both spellings and any depth:

```
plain namespace          ["pkg-a"]
qualified namespace      ["pkg-b"]
qualified deeper         ["pkg-c"]
declare module string    ["pkg-m"]
no block                 []
```

One more test for the qualified case, failing on the previous commit. The tsc extractor specs are 77 passing, and prettier is clean.

sverweij:
Hi @arpitjain099 thanks for addressing it! (Branch) test code coverage seems to be lower than c8 expects; I haven't checked which branches it found uncovered and whether they're part of the contributed code - but in the latter case we might want to address it.

Merging this commit will surface new violations in existing, unmodified code bases. I.a.w. as-is it's a breaking change. There's two ways we can go about it: (1) accept it as a breaking change (2) introduce a feature switch for it. Neither is super-attractive, but as a maintainer I have a slight preference for the second one; we did the same for [dependencies in jsdoc comments](https://github.com/sverweij/dependency-cruiser/blob/main/doc/options-reference.md#detectjsdocimports-detect-dependencies-in-jsdoc-comments) and [dependencies that only exist before compilation](https://github.com/sverweij/dependency-cruiser/blob/main/doc/options-reference.md#tsprecompilationdeps). Reason for my preference is the adoption rate for  minor/ patch bumps is much higher than for major bumps, ensuring more people get the latest features/ fixes (security related or otherwise). For both other examples we added a thing to `--init` so new installs have the new behavior, and existing installs can opt-in (I can contribute the switching functionality if needed).

